PR #9, opened 2018-04-23
RDM-2014 fix for Date field that prevents moving forward

tzarsmango:
due to required validator

https://tools.hmcts.net/jira/browse/RDM-2014

Added a fix for removing validator function in case the component is destroyed (by moving to other pages, etc.)

```
[ ] Yes
[X] No
```


src/components/form/date-input/date-input.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, forwardRef, Input } from '@angular/core';
+import {Component, forwardRef, Input, OnDestroy} from '@angular/core';
```
adr1ancho: Do you need space between { and the next word

PR #13, opened 2018-05-17
RDM 1851: Header and footer needs to match Gov.Uk styling

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-1851

Dependency note: This change has dependency with the change on ccd-case-management-web branch RDM-1851

The header should look like with similar dimensions as in: https://reform-pattern-library.herokuapp.com/
The existing menus should be moved to the black header part, as seen in the pattern library header with menu "Elements""components" etc.

Footer should look like with the similar design as shown in the UI design layout.

Header and Footer will follow the GOV.IN standards in terms of the desktop screen width, etc.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/phase/phase.scss
+++ b/src/components/phase/phase.scss
@@ -47,3 +47,6 @@
 .phase-banner {
    @include phase-banner();
 }
+.full-screen {
+  max-width: 100%;
+}
--- a/src/components/phase/phase.scss
+++ b/src/components/phase/phase.scss
@@ -4,8 +4,10 @@
 
 @mixin phase-banner() {
   padding-top: 10px;
-  padding-left: 22px;
+  padding-left: 15px;
   border-bottom: 1px solid $border-colour;
+  max-width: 1005px;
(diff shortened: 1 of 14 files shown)


src/components/phase/phase.scss
```diff
@@ -4,8 +4,10 @@
 
 @mixin phase-banner() {
   padding-top: 10px;
-  padding-left: 22px;
+  padding-left: 15px;
   border-bottom: 1px solid $border-colour;
```
adr1ancho: Not sure whether this will make code more complicated if we use var() like https://www.w3schools.com/css/css3_variables.asp https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables  Just thought this is now so cumbersome whenever a minor change is required

MKaramath: I have used the vars in most of the places wherever applicable, but in this case, since these styles are already defined, and to reduce complexity i have just modified the px as it is.

PR #19, opened 2018-06-21
RDM-2029:New case payment history list component that will be provide…

andrewfolga:
DO NOT MERGE.

This is for payments team to show how we could integrate their payment history list component.

https://tools.hmcts.net/jira/browse/RDM-2029

### Change description ###
Payment History UI Component

```
[ ] Yes
[x] No
```


src/components/payment/case-payment-history.service.ts
```diff
@@ -0,0 +1,22 @@
+import { Injectable } from '@angular/core';
+import { Payment } from './payment.model';
+import { HttpService } from '../../services/http.service';
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class CasePaymentHistoryService {
```
mario-paniccia: I'm not 100% sure but I don't know whether we want to have Services in the toolkit with complex logic like http calls 
I was under the impression that the toolkit is just for display components. 
Perhaps we can have the service live in our management web and live the component + template here?

andrewfolga: If we have the service live in management web we would still have to pass it to component here, no?

mario-paniccia: yes. The component here will require a service as dependency. Management web will be responsible to provide a service to retrieve payments. 
This way we can keep ui-toolkit simple and only dealing with ui stuff

src/components/payment/context-map.model.ts
```diff
@@ -0,0 +1,24 @@
+export class ContextMap {
```
mario-paniccia: if the ContextMap lives in the toolkit, when in management web we want to pass a new property to a component that lives in management web only, do we have to release a new version of the toolkit?
Is it possible to find a way to have this map live in management web?

PR #41, opened 2018-08-16
RDM-2468: Date field error correction

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2468

### Change description ###
At the moment, when entering a date, a validation error gets displayed as soon as the user leaves the first date input (day) to enter the second date input (month).

```
[ ] Yes
[x] No
```


src/components/form/date-input/date-input.component.ts
```diff
@@ -173,9 +174,23 @@ export class DateInputComponent implements ControlValueAccessor, Validator, OnIn
...
+  }
+
   public inputBlur() {
-    this.formControl.markAsTouched();
-    this.propagateChange(this.rawValue);
+    this.isTouched = true;
+    this.touch();
+  }
+
+  public touch() {
```
tchow8: I don't see touch() being called anywhere in the git diff, so wouldn't:
```
public inputFocus() {
    this.formControl.markAsUntouched();
}

public inputBlur() {
    this.formControl.markAsTouched();
    this.propagateChange(this.rawValue);
}
```
Do the same thing?

MKaramath: touch() is called at Line 184 in date-input.component.ts

src/components/form/date-input/date-input.component.ts
```diff
@@ -173,9 +174,23 @@ export class DateInputComponent implements ControlValueAccessor, Validator, OnIn
...
+    if (this.isTouched) {
```
vlaurin: Instead of having a `isTouched` property and a `touch()` function; I think it would be more explicit to remove the property `isTouched` and split the `touch()` function into 2: `touch()` and `untouch()`.

PR #64, opened 2018-10-03
Rdm 3023 Extract Palette from Case Management web

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3023

### Change description ###
Extract Palette from Case Management web

```
[ ] Yes
[x] No
```


src/app.config.ts
```diff
@@ -0,0 +1,61 @@
+import { CaseEventData } from './shared/domain/case-event-data';
```
andrewfolga: This is an abstract class not an interface as otherwise it would not be possible to use it as an Injection Token in Angular.

This class acts as an interface for all the service that required the real AppConfig.

src/shared/domain/wizard-page.model.ts
```diff
@@ -0,0 +1,35 @@
+import { Orderable } from './order/orderable.model';
+import { WizardPageField } from './wizard-page-field.model';
+import { CaseField } from './definition/case-field.model';
+import { ShowCondition } from '../conditional-show/conditional-show.model';
+import { Type } from 'class-transformer';
+
+// @dynamic
```
andrewfolga: You will see these added in few places. These are to let the compiler know the arrow syntax is acceptable in this class. Otherwise the collector in the compiler was barking at me. :(

src/karma.conf.js
```diff
@@ -17,15 +19,16 @@ module.exports = function (config) {
...
-    reporters: ['progress', 'kjhtml'],
+
+    reporters: ['mocha'],
     port: 9876,
     colors: true,
     logLevel: config.LOG_INFO,
-    autoWatch: true,
-    browsers: ['Chrome'],
-    singleRun: false
+    autoWatch: false,
```
andrewfolga: To follow mgmt web approach.

andrewfolga:
Added some comments on the new changes. All the rest is just moving exactly same code.

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: node_js
 cache: yarn
+install:
+- npm install -g ng-cli
```
andrewfolga: This was needed for tests to start running.

andrewfolga: Actually noticed although it made the build pass green the tests are not running ...

PR #96, opened 2018-10-18
Rdm 3144 Added CaseTypeLite to Jurisdiction

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-3144
Added CaseTypeLite object and referred in Jurisdiction. Jurisdiction used to refer the CaseType which caused cyclic dependency issues on some tests.

```
[ ] Yes
[X] No
```


src/index.ts
```diff
@@ -71,6 +71,7 @@ export { AlertLevel } from './shared/alert/alert-level.model';
 export { Jurisdiction } from './shared/domain/definition/jurisdiction.model';
 export { RouterHelperService } from './shared/utils/router-helper.service';
 export { CaseType } from './shared/domain/definition/case-type.model';
+export { CaseTypeLite } from './shared/domain/definition/case-type-lite.model';
```
mario-paniccia: my only question is why we have CCD domain model classes into the toolkit rather that management web. More a general concern rather than an issue specific with this PR

PR #110, opened 2018-10-31
Rdm 3230 Vanilla app to demonstrate new case progression facade components

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3230

### Change description ###
Case UI Toolkit: Self-document
Changes to library itself:
1. ProfileService moved as required by palette components
2. ProfileNotifier introduced to fix the event trigger retrieval problem
3. LabelFieldComponent added to PaletteModule.
4. Dialogs grouped together and exported via DialogsModule and also CaseUIToolkitModule.
5. Refactor old rxjs 5 syntax to 6: map, throwError, pipe
6. EventTriggerService to announce when event trigger has been loaded from CaseCreateComponent or CaseProgressComponent or from the CaseEditWizardGuard
7. Fix to the Router bug reported here https://github.com/angular/angular/pull/19374
and here https://github.com/angular/angular/pull/26496 that results in it being unable to find component bound to a given navigation path.
8. Fix tests.

```
[x] Yes
[] No
```


demo/README.md
```diff
@@ -0,0 +1,116 @@
+# `angular-library-seed` demo projects
```
andrewfolga: This is not up to date yet.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "1.2.27",
+  "version": "1.2.29",
```
andrewfolga: release new version as changes inside the library were needed as part of this PR (some fixes like missing resolver in CaseEditorModule)

package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
   "version": "2.0.0",
+  "engines": {
+    "yarn": "1.3.2",
```
tzarsmango: Better upgrade this yar engine requirement to the latest stable 1.12.1

andrewfolga: Will do

PR #146, opened 2018-11-22
RDM-3284 CRUD contract on Collection for Add new item

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-3284

This task if for the FE. Checks the ACLs being delivered by the backend see (RDM-2455) to render the add buttons according to the user's rights.

Must be merged after [this PR.](https://github.com/hmcts/ccd-data-store-api/pull/253)

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-form/case-edit-form.html
+++ b/src/shared/components/case-editor/case-edit-form/case-edit-form.html
@@ -1,6 +1,6 @@
 <ng-container *ngFor="let field of fields">
     <div ccdConditionalShow ccdLabelSubstitutor [caseField]="field" [formGroup]="formGroup" [eventFields]="eventFields">
-      <ng-container [ngSwitch]="field | ccdIsReadOnly">
+      <ng-container [ngSwitch]="field | ccdIsReadOnlyAndNotCollection ">
         <ccd-field-read *ngSwitchCase="true" [caseField]="field" [withLabel]="true"></ccd-field-read>
         <ng-container *ngSwitchCase="false">
           <ng-container *ngIf="!(field | ccdIsCompound); else CompoundRow">
--- a/src/shared/components/palette/collection/write-collection-field.component.spec.ts
+++ b/src/shared/components/palette/collection/write-collection-field.component.spec.ts
@@ -3,17 +3,18 @@ import { FormControl, FormGroup, ReactiveFormsModule } from '@angular/forms';
 import { WriteCollectionFieldComponent } from './write-collection-field.component';
 import { DebugElement } from '@angular/core';
 import { MockComponent } from 'ng2-mock-component';
-import { FieldType } from '../../../domain/definition/field-type.model';
-import { CaseField } from '../../../domain/definition/case-field.model';
-import { PaletteUtilsModule } from '../utils/utils.module';
+import { FieldType } from '../../../domain/definition';
+import { CaseField } from '../../../domain/definition';
+import { PaletteUtilsModule } from '../utils';
 import { By } from '@angular/platform-browser';
-import { FormValidatorsService } from '../../../services/form/form-validators.service';
+import { FormValidatorsService } from '../../../services/form';
 import { MatDialog, MatDialogRef } from '@angular/material';
 import { ScrollToService } from '@nicky-lenaers/ngx-scroll-to';
 import { of } from 'rxjs';
-import { RemoveDialogComponent } from '../../dialogs/remove-dialog/remove-dialog.component';
+import { RemoveDialogComponent } from '../../dialogs/remove-dialog';
 import createSpyObj = jasmine.createSpyObj;
 import any = jasmine.any;
+import { ActivatedRoute } from '@angular/router';
 
 const FIELD_ID = 'Values';
 const SIMPLE_FIELD_TYPE: FieldType = {
@@ -71,6 +72,7 @@ describe('WriteCollectionFieldComponent', () => {
   let dialog: any;
   let dialogRef: any;
   let scrollToService: any;
+  let route: any;
 
   let caseField: CaseField;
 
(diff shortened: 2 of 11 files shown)


src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -103,6 +112,35 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
     return index ? `${this.caseField.label} ${displayIndex}` : this.caseField.label;
   }
 
+  isNotAuthorisedToCreate() {
+    if (this.isExpanded) {
+      return false;
+    }
+    return !this.profile.user.idam.roles.find(role => !!this.caseField.acls.find( acl => acl.role === role && acl.create === true));
```
andrewfolga: Can we extract `acl.role === role && acl.create === true` to a separate method like `hasCreateAccess`.

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -103,6 +112,35 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
...
+
+  isNotAuthorisedToUpdate(index: number) {
+    if (this.isExpanded) {
+      return false;
+    }
+    let id = false;
+    if (this.formArray.at(index)) {
+      id = this.formArray.at(index).get('id').value;
+    }
+    return !!id && !this.profile.user.idam.roles.find(role => !!this.caseField.acls.find( acl => acl.role === role && acl.update === true));
```
andrewfolga: Can we extract `acl.role === role && acl.update === true` to a separate method like `hasUpdateAccess`.

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -103,6 +112,35 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
...
+
+  isNotAuthorisedToDelete(index: number) {
+    if (this.isExpanded) {
+      return false;
+    }
+    let id = false;
+    if (this.formArray.at(index)) {
+      id = this.formArray.at(index).get('id').value;
+    }
+    return !!id && !this.profile.user.idam.roles.find(role => !!this.caseField.acls.find( acl => acl.role === role && acl.delete === true));
```
andrewfolga: Can we extract acl.role === role && acl.delete === true to a separate method like hasDeleteAccess.

src/shared/components/case-editor/case-edit-form/case-edit-form.html
```diff
@@ -1,6 +1,6 @@
 <ng-container *ngFor="let field of fields">
     <div ccdConditionalShow ccdLabelSubstitutor [caseField]="field" [formGroup]="formGroup" [eventFields]="eventFields">
-      <ng-container [ngSwitch]="field | ccdIsReadOnly">
+      <ng-container [ngSwitch]="field | ccdIsReadOnlyAndNotCollection ">
```
andrewfolga: Cannot we have `ccdIsReadOnly` in place and a new pile being just `ccdIsNotCollection` and combine the two like `field | ccdIsReadOnly | ccdIsNotCollection`?

andrewfolga: By doing that we would not be repeating `ccdIsReadOnly` inside `ccdIsReadOnlyAndNotCollection` I think.

src/shared/components/palette/collection/write-collection-field.component.spec.ts
```diff
@@ -306,3 +331,240 @@ describe('WriteCollectionFieldComponent', () => {
...
+  }));
+
+  it('should render a row with a read field for each items when user does not have update right', () => {
+    let readFields = de.queryAll($READ_FIELDS);
+    let writeFields = de.queryAll($WRITE_FIELDS);
+
+    expect(writeFields.length).toEqual(0);
+    expect(readFields.length).toEqual(2);
+  });
+});
```
andrewfolga: Shouldn't we have some tests on DELETE right - button disabled/enabled etc.?

src/shared/components/palette/collection/write-collection-field.component.ts
andrewfolga: 👍 for the double bang operators 😄

src/shared/components/palette/collection/write-collection-field.component.spec.ts
```diff
@@ -306,3 +331,240 @@ describe('WriteCollectionFieldComponent', () => {
...
+    fixture.detectChanges();
+  }));
+
+  it('should render a row with a read field for each items when user does not have update right', () => {
+    let readFields = de.queryAll($READ_FIELDS);
+    let writeFields = de.queryAll($WRITE_FIELDS);
+
+    expect(writeFields.length).toEqual(0);
+    expect(readFields.length).toEqual(2);
+  });
```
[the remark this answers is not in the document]
tzarsmango: actually we're replacing the input fields (ccd-field-write) with ccd-field-read. So there's no input box if user doesn't have e update right

src/shared/components/case-editor/case-edit-form/case-edit-form.html
tzarsmango: Even though the field is marked as READONLY on the BE (since the user doesn't have UPDATE right), we want to handle COLLECTION fields differently and let the user create new collection items if the user has CREATE right. So on this step we don't want to treat collection as a `<ccd-field-read>` even though the field is set as READ only by the BE.

andrewfolga: Ok all clear now.

PR #147, opened 2018-11-26
RDM-3325: Broken regex fixed; whitespace validation fixed;

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-3325

### Change description ###
Fixed the broken regex and whitespace validation

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/palette/case-link/read-case-link-field.component.ts
+++ b/src/shared/components/palette/case-link/read-case-link-field.component.ts
@@ -6,4 +6,12 @@ import { AbstractFieldReadComponent } from '../base-field/abstract-field-read.co
   templateUrl: 'read-case-link-field.html'
 })
 export class ReadCaseLinkFieldComponent extends AbstractFieldReadComponent {
+
+  getCaseReference(): string {
+    if (this.caseField && this.caseField.value) {
+      return this.caseField.value.CaseReference;
+    }
+    return null;
+  }
+
 }
--- a/src/shared/services/form/form-validators.service.spec.ts
+++ b/src/shared/services/form/form-validators.service.spec.ts
@@ -24,4 +24,15 @@ describe('FormValidatorsService', () => {
     result.updateValueAndValidity();
     expect(result.invalid).toBeTruthy();
   });
+
+  it('should validate text field for MANDATORY with regular expression', () => {
+    let formControl: FormControl = new FormControl();
+    let caseField: CaseField = aCaseField('id', 'label', 'Text', 'MANDATORY', null);
+    caseField.field_type.regular_expression = '^(Test)$'
+    let result: FormControl = formValidatorsService.addValidators(caseField, formControl);
+    result.setValue('Test');
+    result.markAsTouched();
+    result.updateValueAndValidity();
+    expect(result.invalid).toBeFalsy();
+  });
 });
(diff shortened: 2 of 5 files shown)


src/shared/services/form/form-validators.service.spec.ts
```diff
@@ -24,4 +24,15 @@ describe('FormValidatorsService', () => {
     result.updateValueAndValidity();
     expect(result.invalid).toBeTruthy();
   });
+
+  it('should validate text field for MANDATORY with regular expression', () => {
```
andrewfolga: Shouldn't we have tests that would verify what is asked in the ticket? This would be a field with value containing spaces?

Also a review of all tests according to the rules that are expected to be preserved would also be useful.

MKaramath: Done.

andrewfolga: Actually I am not sure that it works properly. Just ran a test like this one:

```
  fit('should validate text field for MANDATORY with regular expression', () => {
    let formControl: FormControl = new FormControl();
    let caseField: CaseField = aCaseField('id', 'label', 'Text', 'MANDATORY', null);
    caseField.field_type.regular_expression = '^([Pp][Bb][Aa])\d{7}$'
    let result: FormControl = formValidatorsService.addValidators(caseField, formControl);
    
   
    result.setValue('PBA 1234567');
    result.markAsTouched();
    result.updateValueAndValidity();
    expect(result.invalid).toBeTruthy();
  });
```

which should pass but is failing. The space should make the test fail as far as I understood the ACs?

src/shared/components/markdown/markdown.component.ts
```diff
@@ -1,11 +1,15 @@
...
 })
-export class MarkdownComponent {
+export class MarkdownComponent implements OnInit {
 
-    @Input()
-    content: string;
+  @Input()
+  content: string;
+
+  ngOnInit(): void {
```
andrewfolga: Should this be part of this story?

src/shared/services/form/form-validators.service.spec.ts
MKaramath: As discussed.

src/shared/components/markdown/markdown.component.ts
MKaramath: This is part of RDM-2974

src/shared/services/form/form-value.service.ts
```diff
@@ -55,8 +55,13 @@ export class FormValueService {
     switch (typeof rawValue) {
       case 'object':
         return this.sanitiseObject(rawValue);
-      case 'string':
-        return rawValue.trim();
+      case 'string': {
+        if (rawValue.trim().length === 0) {
```
andrewfolga: Isn't it equivalent to performing trim in all cases?

src/shared/components/palette/case-link/read-case-link-field.component.ts
```diff
@@ -6,4 +6,12 @@ import { AbstractFieldReadComponent } from '../base-field/abstract-field-read.co
   templateUrl: 'read-case-link-field.html'
 })
 export class ReadCaseLinkFieldComponent extends AbstractFieldReadComponent {
+
+  getCaseReference(): string {
+    if (this.caseField && this.caseField.value) {
+      return this.caseField.value.CaseReference;
```
mario-paniccia: Karamath would this break with Draft cases that have no reference?

PR #154, opened 2018-11-28
 RDM-3463: Update dependency on compodoc to Version 1.1.7

Daniel-Lam:
https://tools.hmcts.net/jira/browse/RDM-3463.

### Change description ###
Fixes vulnerability where `compodoc` is dependent on a version of `live-server` that is dependent on a compromised version of `event-stream`.

```
[ ] Yes
[x] No
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ docs/
 # demo
 demo/dist/
 demo/node_modules
+
+# package-lock
+package-lock.json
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -51,3 +51,6 @@ docs/
 # demo
 demo/dist/
 demo/node_modules
+
+# package-lock
```
andrewfolga: it's ok but shouldn't we be building it with yarn thus no need for package-lock.json?

Daniel-Lam: Yep - that's correct. I built it with `yarn` but IntelliJ IDEA generated a `package-lock.json`, so I added the exclusion just in case

andrewfolga: it is fine then

PR #162, opened 2018-12-04
RDM-2963 changed initialisation values to null 

tzarsmango:
changed initialisation values to null for YesNo, fixedList and fixedRadioList components

The change was requested due to issues encountered in the elastic search queries

https://tools.hmcts.net/jira/browse/RDM-2963

```
[ ] Yes
[X] No
```


src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -75,4 +75,9 @@ export class ShowCondition {
   matchByCaseFields(caseFields: CaseField[]): boolean {
     return this.match(FieldsUtils.toValuesMap(caseFields));
   }
+
+  private okIfBothEmpty(right: string, value: any) {
+    return value === null && (right === '');
```
andrewfolga: Can right be empty string if we are setting it to null everywhere?

PR #165, opened 2018-12-06
RDM-3418 Validation errors prevents navigation

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-3418

Removing the code piece that prevents moving forward in case of validation errors

```
[ ] Yes
[X] No
```


tzarsmango:
This does not change the way warnings/errors. They are displayed in banner and around the failing control as usual. It just removes the code that causes errors preventing to move forward in any way. However a general regressions is advised.

andrewfolga:
What I meant here is I think there was a requirement long time ago to disable the button if there were errors/warnings from callbacks. Just to double check such func is not expected etc.

PR #192, opened 2019-01-16
Revert "Revert "RDM-3284 CRUD contract on Collection for Add new item""

tzarsmango:
Reverts hmcts/ccd-case-ui-toolkit#189


src/shared/components/case-editor/case-edit-page/case-edit-page.component.spec.ts
```diff
@@ -544,7 +549,7 @@ describe('CaseEditPageComponent', () => {
       fixture.detectChanges();
     });
 
-    it('should call validate', async () => {
+    it('should call validateee', async () => {
```
andrewfolga: why the name change :)

andrewfolga: This is a tiny rename to fix the test name :)

PR #196, opened 2019-01-28
RDM-3799 New contract for case view component.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3799

### Change description ###
New contract for case view component plus example usage in heroku demo app.

```
[ ] Yes
[x] No
```


demo/src/app/case-view-consumer.component.ts
```diff
@@ -0,0 +1,20 @@
+import { Component, OnInit } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
+
+@Component({
+  selector: 'case-create-consumer',
+  template: `<div class="container-fluid">
+                <ccd-case-view [case]="caseId"></ccd-case-view>
+             </div>`
+})
+export class CaseViewConsumerComponent implements OnInit {
```
gbenadikar: should the name be CaseCreateConsumerComponent?

demo/src/app/case-view-consumer.component.ts
```diff
@@ -0,0 +1,20 @@
...
+    caseId;
```
gbenadikar: type for caseId?

demo/src/app/case-view-consumer.component.ts
andrewfolga: I think this name is actually correct as it is consistent with the other consumers like CaseCreateConsumerComponent or CaseProgressConsumerComponent.

demo/src/app/case-view-consumer.component.ts
andrewfolga: Added.

PR #198, opened 2019-02-01
Rdm 3821 Extract Case Search Form to ui-toolkit as is

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3821

### Change description ###
Extract Case Search Form to ui-toolkit as is

```
[ ] Yes
[x] No
```


src/shared/components/search-filters/search-filters.component.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  constructor(private searchService: SearchService,
+    private orderService: OrderService,
+    private jurisdictionService: JurisdictionService,
+    private windowService: WindowService) {
+  }
+
+  ngOnInit(): void {
+    this.selected = {};
+    const jurisdiction = this.windowService.getLocalStorage(JURISDICTION_LOC_STORAGE);
```
mohammedanas: Extracting Search component into CaseUIToolKit also includes LocalStorage persistent functionality, so other parties other than case management using this component may get this feature even if they haven't asked for or even if they don't want it. But I guess the business might be able to answer it more appropriately.

Let me know your thoughts

andrewfolga: Good point and as discussed this will be more like a feature of the existing extracted component to third parties so it is ok.

PR #210, opened 2019-02-07
Rdm 3675 doc link

andrewfolga:
A fix to image sourcing. It has to be defined as close to the component as possible and not in the library clients.

```
[ ] Yes
[x] No
```


src/shared/components/case-viewer/event-log/event-log-table.component.ts
```diff
@@ -1,7 +1,5 @@
 import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { CaseViewEvent } from '../../../domain';
-import { ActivatedRoute } from '@angular/router';
-import { CasesService } from '../../case-editor';
```
andrewfolga: Unused imports.

PR #212, opened 2019-02-11
Rdm 3830

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3830

### Change description ###
Add new Search Filters component to demo app.

```
[ ] Yes
[x] No
```


demo/server.js
```diff
@@ -34,6 +34,12 @@ app.use(function (req, res) {
         response = db.get('case').value();
       } else if (req.url.indexOf('/documents') !== -1) {
         response = db.get('documents').value();
+      } else if (req.url.indexOf('CT0/inputs') !== -1) {
+        response = db.get('CT0inputs').value();
+      } else if (req.url.indexOf('CT2/inputs') !== -1) {
+        response = db.get('CT2inputs').value();
+      } else if (req.url.indexOf('CT3/inputs') !== -1) {
+        response = db.get('CT3inputs').value();
```
mohammedanas: Would it be a reasonable case for refactoring it to a switch, I understand its an update to existing piece of code, I do not mind if you want to leave it as it is, just a thought.

andrewfolga: Not really here as there is a substring search there rather than exact match.

PR #218, opened 2019-02-12
Feature: Rdm 3915 case filter as is

adnanakgun:
https://tools.hmcts.net/jira/browse/RDM-3915

### Change description ###
Extracted component "as is" from ccd-case-management-web

```
[ ] Yes
[x] No
```


demo/src/app/case-create-filter.component..ts
```diff
@@ -0,0 +1,13 @@
+import { Component } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
+
+@Component({
+  selector: 'case-create-filter',
+  template: `<div class="container-fluid">
+                <ccd-create-case-filters></ccd-create-case-filters>
+             </div>`
+})
+export class CaseCreateFilterComponent {
```
andrewfolga: Can we rename to CaseCreateFilterConsumerComponent here to make it consistent with the others?

demo/src/app/app.routing.ts
```diff
@@ -24,6 +25,9 @@ export const routing = RouterModule.forRoot([
       { path: 'case/:jid/:ctid/:cid',
         component: CaseViewConsumerComponent,
         children: viewerRouting
+      },
```
andrewfolga: This is part of a future task in this epic: https://tools.hmcts.net/jira/browse/RDM-3918

demo/src/app/app.module.ts
```diff
@@ -37,6 +38,7 @@ import { CaseViewConsumerComponent } from './case-view-consumer.component';
     CaseCreateConsumerComponent,
     CaseProgressConsumerComponent,
     CaseViewConsumerComponent,
+    CaseCreateFilterComponent
```
andrewfolga: This is part of a future task in this epic: https://tools.hmcts.net/jira/browse/RDM-3918

demo/server.js
```diff
@@ -34,6 +34,8 @@ app.use(function (req, res) {
         response = db.get('case').value();
       } else if (req.url.indexOf('/documents') !== -1) {
         response = db.get('documents').value();
+      }else if (req.url.indexOf('aggregated/caseworkers/:uid/jurisdictions') !== -1) {
```
andrewfolga: This is part of a future task in this epic: https://tools.hmcts.net/jira/browse/RDM-3918

demo/src/app/core.component.ts
```diff
@@ -14,6 +14,9 @@ import { Component } from '@angular/core';
                         <li>
                         <a [routerLink]="'/case'" [routerLinkActive]="'item-bold'">View Case</a>
                         </li>
+                        <li>
+                        <a [routerLink]="'/case/filters'" [routerLinkActive]="'item-bold'">Case Create Filters</a>
```
andrewfolga: This is part of a future task in this epic: https://tools.hmcts.net/jira/browse/RDM-3918

demo/stubs/api.json
```diff
@@ -3574,5 +3574,4415 @@
        "type":"CCD Print Type",
        "description":"Printing for CCD"
     }
+ ],
+ "jurisdictions": [
```
andrewfolga: This is part of a future task in this epic: https://tools.hmcts.net/jira/browse/RDM-3918

src/shared/components/case-editor/case-editor.module.ts
```diff
@@ -53,6 +55,7 @@ import { ErrorsModule } from '../error/errors.module';
         CaseEditFormComponent,
         CaseEditSubmitComponent,
         CaseCreateComponent,
+        CreateCaseFiltersComponent,
```
andrewfolga: I do not think we need this import here. It won't be used in this module.

demo/src/app/case-create-filter.component..ts
```diff
@@ -0,0 +1,13 @@
+import { Component } from '@angular/core';
```
andrewfolga: file name has two dots

src/shared/components/case-editor/case-editor.module.ts
andrewfolga: We should have a new CreateCaseFiltersModule with all necessary providers declaring one CreateCaseFilters component and exporting.

demo/src/app/case-create-filter.component..ts
adnanakgun: Done

demo/src/app/app.routing.ts
adnanakgun: Seem to have highlighted wrong section, but I get it and have removed that bit.

demo/src/app/app.module.ts
adnanakgun: Done

demo/server.js
adnanakgun: Done

demo/src/app/core.component.ts
adnanakgun: Done

demo/stubs/api.json
adnanakgun: Done

demo/src/app/case-create-filter.component..ts
adnanakgun: File removed

src/shared/components/case-editor/case-editor.module.ts
adnanakgun: I will do as suggested and do the "ccd-case-management-web" PR as part of this change. Therefore, will need a little more time.

src/case-ui-toolkit.module.ts
```diff
@@ -71,6 +72,7 @@ import { PipesModule, CaseReferencePipe, MarkdownComponent, MarkdownModule, Base
       CaseViewComponent,
       CaseViewerComponent,
       SearchFiltersComponent,
+      CreateCaseFiltersComponent,
```
andrewfolga: Historically we used to import every newly migrated component into CaseUIToolkitModule. This approach was there because initially this repo was a very small library of components: see /src/components and we used to need all of them so grouping made sense. It has grown in size considerably since then.

I realised we are importing a lot of unnecessary components as part of it in man web whenever we import it like AlertModule.

We should have a module per component approach and import only the necessary modules. Also future clients should not need to import a one module with all components in it when they will need just one or two.

Also see here: https://stackoverflow.com/questions/46434830/should-i-create-a-module-per-component-in-angular-4-app.

I will follow that from now on as well. Feel free to ask questions.

andrewfolga:
I would keep the demo app changes in here - no point in removing them just to re-add them later 👍

src/case-ui-toolkit.module.ts
adnanakgun: I've taken this out now.

src/shared/components/case-editor/case-editor.module.ts
adnanakgun: Both PR's are now in place.

PR #227, opened 2019-02-20
Rdm 3897- Add optional property casereference to the CaseEventData object

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-3897
As a MidEvent Callback

I want to retrieve all the case data, previous page data entered in the current event and current page data

When the event the mid event triggered

```
[ ] Yes
[X] No
```


src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -128,6 +128,10 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       caseEventData.ignore_warning = this.ignoreWarning;
       caseEventData.event_data = this.editForm.value.data;
 
+      if (this.caseEdit.caseDetails) {
```
gbenadikar: Can this be covered by an unit test? Any expect in existing test should suffice

src/shared/domain/case-event-data.model.ts
```diff
@@ -9,4 +9,5 @@ export class CaseEventData {
   event_token: string;
   ignore_warning: boolean;
   draft_id?: string;
+  case_reference?: string;
```
mario-paniccia: Shouldn't case reference be mandatory? i.e. `case_reference` rather than `case_reference?`

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
hemantt: @mohammedanas

PR #237, opened 2019-02-26
RDM-3940 Add Hide and Show support to new ComplexTypeToEvent

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-3940

### Change description ###
Adds new show and hide functionality for ComplexTypeToEvent mapping.
When creating the case eventFields and idPrefix are passed to ConditionalShowDirective to be able to show or hide the field in a complex field based on other event fields data. idPrefix is used to determine which collection element are we in.

```
[ ] Yes
[ X] No
```


src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -122,7 +125,7 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
   }
 
   buildIdPrefix(field: CaseField): string {
-    return this.isCompoundPipe.transform(field) ? `${this.idPrefix}${field.id}_` : `${this.idPrefix}`;
+    return this.isCompoundPipe.transform(field) ? `${this.idPrefix}_` : `${this.idPrefix}`;
```
gbenadikar: 👍

src/shared/components/palette/base-field/field-read.component.ts
```diff
@@ -33,6 +41,11 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
 
     // Provide component @Inputs
     component.instance['caseField'] = this.caseField;
+    // console.log('this.caseField', this.caseField);
```
gbenadikar: Can we remove these commented console.logs?

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -70,8 +76,10 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
 
   buildIdPrefix(index: number): string {
     if ('Complex' === this.caseField.field_type.collection_field_type.type) {
+      // console.log('collection idPrefix.id', this.idPrefix + this.caseField.id + '_' + index + '_');
```
gbenadikar: remove console.logs?

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -10,24 +10,28 @@ export class ShowCondition {
   constructor(public condition: string) {
   }
 
-  match(fields): boolean {
+  match(fields, path?: string): boolean {
+    // if (path) {
```
gbenadikar: can we remove this block if not required?

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -56,11 +60,34 @@ export class ShowCondition {
...
 
-  private findValueForComplexCondition(fields: any, head: string, tail: string[]) {
+  private findValueForComplexCondition(fields: any, head: string, tail: string[], path?: string) {
     if (tail.length === 0) {
       return fields[head];
     } else {
-      return this.findValueForComplexCondition(fields[head], tail[0], tail.slice(1));
+      if (Array.isArray(fields[head])) {
+        // use the path to resolve which array element we refer to
+        if (path.startsWith(head)) {
```
gbenadikar: will it be better to create a separate function for this if block (if logical to do so)?

src/shared/directives/conditional-show/domain/conditional-show.model.spec.ts
```diff
@@ -303,6 +357,37 @@ describe('conditional-show', () => {
 
       expect(matched).toBe(false);
     });
+
+    it('should return false when value will not match on a collection element with a complex field', () => {
```
gbenadikar: 👍

src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -81,8 +85,7 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
   }
 
   isComplexWithHiddenFields() {
-    if (this.caseField.field_type
-      && this.caseField.field_type.complex_fields
+    if (this.caseField.field_type.type === 'Complex' && this.caseField.field_type.complex_fields
```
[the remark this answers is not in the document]
rafalkalita: caseField.field_type is defined as mandatory, so the check is not needed.

src/shared/components/palette/base-field/field-read.component.ts
rafalkalita: Will remove all of them.

src/shared/directives/conditional-show/domain/conditional-show.model.ts
rafalkalita: I will leave it this way, because IDE doesn't want to do it for me :)

src/shared/components/case-editor/case-edit-form/case-edit-form.html
```diff
@@ -1,15 +1,17 @@
 <ng-container *ngFor="let field of fields">
     <div ccdConditionalShow ccdLabelSubstitutor [caseField]="field" [formGroup]="formGroup" [eventFields]="eventFields">
       <ng-container [ngSwitch]="field | ccdIsReadOnlyAndNotCollection ">
-        <ccd-field-read *ngSwitchCase="true" [caseField]="field" [withLabel]="true"></ccd-field-read>
+        <ccd-field-read *ngSwitchCase="true" [caseField]="field" [eventFields]="eventFields" [withLabel]="true"></ccd-field-read>
```
andrewfolga: Just initial thought - why are we passing this here? I worked on similar story for making case history a flexible field and Fatih and I agreed we should not be cluttering ccd-field-read with eventFields therefore moved the logic to the BE. Cannot we do that here as well?

andrewfolga: Can you let me know why do you need eventFields so we can think together how to better design it?

andrewfolga: I think the initial name for the `eventFields` @Input was chosen because the ConditionalShowDirective was only to be used on case progression. I think it could be renamed now to `caseFields` as it creates confusion if the directive is used in the case view context. Let me know what you think?

src/shared/components/palette/address/write-address-field.component.ts
andrewfolga: I wonder why did we have the change in the first place? 😄
Talking about `${this.idPrefix}${field.id}_`

src/shared/components/case-editor/case-edit-form/case-edit-form.html
rafalkalita: I already wanted to name it caseFields, but was worried it will not be backwards compatible if someone else is using the component already and we change the Input field name.

src/shared/directives/conditional-show/conditional-show.directive.ts
```diff
@@ -31,18 +32,16 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
   ngAfterViewInit() {
     if (this.caseField.show_condition) {
       this.condition = new ShowCondition(this.caseField.show_condition);
```
andrewfolga: You may want to check with mario but he was always asking to keep those console.log statements there for future debugging.

src/shared/directives/conditional-show/conditional-show.directive.ts
```diff
@@ -17,7 +17,8 @@ import { ConditionalShowRegistrarService } from './services/conditional-show-reg
 export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
 
   @Input() caseField: CaseField;
-  @Input() eventFields: CaseField[] = [];
+  @Input() idPrefix: string;
+  @Input() eventFields: CaseField[] = []; // all case data - should be caseFields, but keeping old name for backward compatibility
```
andrewfolga: Can we change it to caseFields?

src/shared/directives/conditional-show/conditional-show.directive.ts
```diff
@@ -31,18 +32,16 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
   ngAfterViewInit() {
     if (this.caseField.show_condition) {
       this.condition = new ShowCondition(this.caseField.show_condition);
-      // console.log('FIELD: ' + this.caseField.id + ' init. Show condition: ' + this.caseField.show_condition);
       this.formGroup = this.formGroup || new FormGroup({});
       this.formField = this.formGroup.get(this.caseField.id);
-      // console.log('FIELD: ' + this.caseField.id + '. Is form field:' + this.formField + '. Event fields:', this.eventFields);
+
```
andrewfolga: Similar to above about Mario's wanting to leave it.

src/shared/directives/conditional-show/conditional-show.directive.ts
```diff
@@ -31,18 +32,16 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
...
       this.updateVisibility(this.getReadOnlyAndFormFields());
       this.subscribeToFormChanges();
       this.registry.register(this);
     }
   }
 
   refreshVisibility() {
-    // console.log('Refresh FIELD: ', this.caseField.id, '. field:', this.formField, '. eventFields:', this.eventFields);
```
andrewfolga: Same to all below ones 😄

src/shared/directives/conditional-show/conditional-show.directive.ts
gbenadikar: well commented lines of code don't look great and leaving them so that the developers can uncomment and debug seems a bit lazy 😁

andrewfolga: I fully agree 😄 Please ask Mario why he wanted that in. I wanted to remove it too last time.

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -10,24 +10,24 @@ export class ShowCondition {
...
 
-  private matchAndConditions(fields: any, condition: string): boolean {
+  private matchAndConditions(fields: any, condition: string, path?: string): boolean {
     let andConditions = condition.split(ShowCondition.AND_CONDITION_REGEXP);
-    return andConditions.every(andCondition => this.matchEqualityCondition(fields, andCondition));
+    return andConditions.every(andCondition => this.matchEqualityCondition(fields, andCondition, path));
   }
 
-  private matchEqualityCondition(fields: any, condition: string): boolean {
+  private matchEqualityCondition(fields: any, condition: string, path?: string): boolean {
```
andrewfolga: This method is a bit bulky. Is it possible to break it down a bit?

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -56,11 +56,34 @@ export class ShowCondition {
     }
   }
 
-  private findValueForComplexCondition(fields: any, head: string, tail: string[]) {
+  private findValueForComplexCondition(fields: any, head: string, tail: string[], path?: string) {
```
andrewfolga: This method is a bit bulky. Is it possible to break it down a bit?

src/shared/directives/conditional-show/conditional-show.directive.ts
mario-paniccia: @andrewfolga is right. Those comments are there for a reason. Hide/show is quite tricky, and investigating issues if complicated without some good logging. That's why we 
decided to leave them there. At least, until we find a better approach. But for the time being, please put them back

src/shared/components/palette/address/write-address-field.component.ts
rafalkalita: it is because for complex types idPrefix had duplicates, for example interimReturns_addressAttended_addressAttended_PostCode
As well I couldn't find any usage of idPrefix in the current codebase! Neither in management_web

src/shared/directives/conditional-show/conditional-show.directive.ts
rafalkalita: Wanted to say we should use console.debug, but in JavaScript it will just change the clour of the message. Have no counter arguments. Will put the comments back :)

src/shared/directives/conditional-show/conditional-show.directive.ts
rafalkalita: done

src/shared/components/case-editor/case-edit-form/case-edit-form.html
rafalkalita: Confirmed with Andrew that this change will not break the contract. Changed eventFields to caseFields.

src/shared/components/palette/base-field/field-read.component.ts
```diff
@@ -19,6 +21,12 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
   @Input()
   withLabel = false;
 
+  @Input()
```
mario-paniccia: any field that is inject by this class to the various field read components, (`component.instance['caseReference'] = this.caseReference;`), is better declared on the abstract class AbstractFieldReadComponent. If you look at the fields that are currently injected, i.e. `caseField`, `caseReference`, `context`, these are all declared once on `AbstractFieldReadComponent`

mario-paniccia: On another note, can I know why these two fields have been added to this component?

src/shared/components/palette/base-field/field-write.component.ts
```diff
@@ -26,11 +26,14 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements
   @Input()
   formGroup: FormGroup;
 
+  @Input()
+  caseFields: CaseField[] = [];
```
mario-paniccia: same comment as the one added to field-read.component.ts. `caseFields` should be declared in `AbstractFieldWriteComponent`

src/shared/directives/conditional-show/conditional-show.directive.ts
```diff
@@ -17,7 +17,8 @@ import { ConditionalShowRegistrarService } from './services/conditional-show-reg
 export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
 
   @Input() caseField: CaseField;
-  @Input() eventFields: CaseField[] = [];
+  @Input() idPrefix: string;
+  @Input() caseFields: CaseField[] = [];
```
mario-paniccia: 👍  for renaming, but the new name is not entirely correct. `conditional-show.directive.ts` can be applied in different contexts. For example to tabs, and in that case the `caseFields` array will only contain the fields declared in the tab. I think something like `contextFields` might be more appropriate.

src/shared/directives/conditional-show/domain/conditional-show.model.ts
mario-paniccia: I agree with @andrewfolga, the complexity of this class has increased a lot. We need refactoring and further simplification. It's a bit difficult to follow

src/shared/directives/conditional-show/domain/conditional-show.model.ts
mario-paniccia: Agree with @andrewfolga

src/shared/components/palette/base-field/field-read.component.ts
rafalkalita: Those were added because ccdConditionalShow and ccdLabelSubstitutor needs them. So if you want to use show and hide or markdown functionality in any template you have to have this data present.

mario-paniccia: Ok so in that case I think we should try to find a better approach to get the case fields and form group down to the ccdConditionalShow and ccdLabelSubstitutor, where they will be actually used. Because adding these fields to the read and write components just to propagate them down could be avoided. These fields does not really belong to these classes. Let's see if we can find a better way to do that

src/shared/components/case-editor/case-edit-submit/case-edit-submit.component.ts
```diff
@@ -131,13 +131,12 @@ export class CaseEditSubmitComponent implements OnInit {
...
 
   summaryCaseField(field: CaseField): CaseField {
-    let cloneField: CaseField = Object.assign({}, field);
-
     if (null == this.editForm.get('data').get(field.id)) {
       // If not in form, return field itself
       return field;
     }
 
+    let cloneField: CaseField = Object.assign({}, field);
```
mario-paniccia: we have a method `cloneObject` on FieldsUtils which this class is already injected with. You could make that method public and reuse it (I think to remember it was public for reusability, not sure why is private now)

src/shared/components/palette/base-field/field-read.component.ts
mario-paniccia: @rafalkalita I gave it a tought but can't find a better solution at the moment. If something comes to my mind later we can refactor as part of a new story, so please ignore the last comment above. But we could move those fields to the abstract classes. That would remove some duplication

src/shared/components/case-editor/services/wizard-page-field-to-case-field.mapper.ts
```diff
@@ -64,6 +66,33 @@ export class WizardPageFieldToCaseFieldMapper {
     }
   }
 
+  private fixShowConditionPath(caseField: CaseField, pathPrefix: string) {
+    if (caseField.show_condition && pathPrefix.length > 0 && !caseField.show_condition.startsWith(pathPrefix)) {
+      caseField.show_condition = pathPrefix.length > 0 ? pathPrefix + '.' + caseField.show_condition : caseField.show_condition;
```
gbenadikar: do we need 'pathPrefix.length > 0' check again as same check is done in the if block?

src/shared/components/case-editor/services/wizard-page-field-to-case-field.mapper.ts
```diff
@@ -64,6 +66,33 @@ export class WizardPageFieldToCaseFieldMapper {
...
+    }
+
+    let childrenCaseFields = [];
+    if (this.isCollection(caseField)) {
```
gbenadikar: These 5 lines can be converted into a function to return child complex fields. I see them repeated in hideParentIfAllChildrenHidden.

src/shared/components/case-editor/services/wizard-page-field-to-case-field.mapper.ts
```diff
@@ -82,7 +84,34 @@ export class WizardPageFieldToCaseFieldMapper {
...
+
+  private allCaseFieldsHidden(children: CaseField[]): boolean {
+    return !children.some(e => e.hidden !== true);
+  }
+
   private isCollectionOfComplex(case_field: CaseField) {
     return case_field.field_type.type === 'Collection' && case_field.field_type.collection_field_type.type === 'Complex';
   }
+
+  private isComplex(case_field: CaseField) {
```
gbenadikar: Would have been good if these functions were on the model itself, if not at least in FieldsUtils to make them reusable.

rafalkalita: Utility methods are on https://tools.hmcts.net/jira/browse/RDM-4073 waiting to be merged

src/shared/components/case-editor/services/wizard-page-field-to-case-field.mapper.ts
rafalkalita: 👍

src/shared/components/case-editor/services/wizard-page-field-to-case-field.mapper.ts
rafalkalita: 👍

gbenadikar:
Last 2 commits look good

src/shared/components/case-editor/services/wizard-page-field-to-case-field.mapper.ts
```diff
@@ -40,9 +40,11 @@ export class WizardPageFieldToCaseFieldMapper {
     const caseFieldIds = override.complex_field_element_id.split('.');
     let case_field_leaf: CaseField;
 
-    if (this.isCollectionOfComplex(caseField)) {
+    let children = this.getCaseFieldChildren(caseField);
```
tzarsmango: should we consider removing the method isCollectionOfComplex since it's not being used anymore?

rafalkalita: Done

PR #246, opened 2019-03-01
RDM-4073 add utility methods to CaseField 

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-4073

Added utility methods `isComplex` and `isCollection` to CaseField. This required many changes to the test code due to the method used for CaseField creation in tests.

```
[ ] Yes
[X] No
```


src/shared/domain/definition/case-field.model.ts
```diff
@@ -25,4 +25,12 @@ export class CaseField implements Orderable {
...
   @Type(() => WizardPageField)
   wizardProps?: WizardPageField;
+
+  isCollection(): boolean {
+    return this.field_type && this.field_type.type === 'Collection';
+  }
+
+  isComplex(): boolean {
+    return this.field_type && this.field_type.type === 'Complex';
+  }
```
mario-paniccia: @tzarsmango I know is not part of the ticket, but I was thinking whether we could also add other methods like isLabel() (I think to remember that's another one very often used:     return field.field_type.type === 'Label';
) and some others you can think of
Feel free to ignore

src/shared/fixture/case-field-builder.ts
```diff
@@ -0,0 +1,74 @@
+import { CaseField, FieldType } from '../domain/definition';
+import { AccessControlList } from '../domain/definition/access-control-list.model';
+
+export class CaseFieldBuilder {
```
andrewfolga: 👍

src/shared/fixture/case-field.test.fixture.ts
```diff
@@ -0,0 +1,19 @@
+import { FieldType } from '../domain/definition';
+import { textFieldType } from '.';
+import { CaseFieldBuilder } from './case-field-builder';
+
+export let newCaseField = (id: string,
```
andrewfolga: 👍

src/shared/domain/definition/case-field.model.ts
mario-paniccia: and a method
Also a method:
getComplexSubfield(path) that allows retrieval of a Complex field subfield by its path.
This logic currently sits in  wizard-page-field-to-case-field.mapper.ts and could be moved in CaseField.

@tzarsmango if you decide to not tackle these as part of this story, please let me know I will create a separate ticket

PR #258, opened 2019-03-06
Added validation to the document component

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-3480
The Document upload component does not validate is left untouched in MANDATORY mode, when its part of complex type.

```
[ ] Yes
[X] No
```


src/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -43,16 +43,22 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
   ngOnInit() {
     this.initDialog();
     let document = this.caseField.value;
+    console.log('this.caseField', this.caseField);
```
rafalkalita: Please remove the logging

src/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -43,16 +43,21 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
   ngOnInit() {
     this.initDialog();
     let document = this.caseField.value;
+    if (this.caseField.display_context && this.caseField.display_context === this.MANDATORY) {
```
mario-paniccia: Mohammed this solution looks ok. 
But can I ask you to make sure we follow the same process to do Mandatory field checks as we do for other fields? If you do a search on the code base for 'MANDATORY' you will see this is already used elsewhere. 
Let's just make sure we use one solution for all our fields. This makes the code more maintainable

mohammedanas: Yes you are right, I tried importing that but, the whole testing and mocking was getting way too complicated with it, I can't remember exactly what it was, but that was the reason it seemed more easier to use the String constant MANDATORY, I shall revisit it and let you know

mario-paniccia: yes please @mohammedanas

PR #270, opened 2019-03-12
RDM-876-1: Hide Print button when not configured

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-876

### Change description ###
As a court staff I do not want to see the Print button if the print callback url is not configured so that I do not have any confusion.

ccd-case-management-web PR (related):
https://github.com/hmcts/ccd-case-management-web/pull/590

ccd-data-store-api PR (related):
https://github.com/hmcts/ccd-data-store-api/pull/401

```
[ ] Yes
[x] No
```


demo/src/app/search-filters-consumer.component.ts
```diff
@@ -15,7 +15,8 @@ export class SearchFiltersConsumerComponent implements OnInit {
       states: [],
       events: [],
       case_fields: [],
-      jurisdiction: null
+      jurisdiction: null,
+     printableDocumentUrl: 'print-url'
```
MSancaktutar: My only comment is to confirm correctness of or fix the singularity of this field.

MKaramath: Done.   Resolved after making the field as optional.

MKaramath:
> this implementation is ok. But considering that the print url is not used by the UI, it would perhaps have been better the backend not passing the url. Instead the backend could just have passed a boolean property called 'isPrintEnabled'.

Done.

MKaramath:
> this implementation is ok. But considering that the print url is not used by the UI, it would perhaps have been better the backend not passing the url. Instead the backend could just have passed a boolean property called 'isPrintEnabled'.

Done.

veeresha-hmcts:
Verified locally and ready for merge.

PR #278, opened 2019-03-15
RDM-4386 Make print, event selector and history tab optional but pres…

andrewfolga:
…ent by default.

https://tools.hmcts.net/jira/browse/RDM-4386

### Change description ###
Make print, event selector and history tab optional but present by default. This is desired for consumers of extracted component and configured in demo app to hide all three mentioned components. Man web will not be affected but regression about visibility of Print, CaseHistory and EventSelector is needed.

```
[ ] Yes
[x] No
```


src/shared/components/case-viewer/case-view/case-view.component.ts
```diff
@@ -14,7 +14,16 @@ export class CaseViewComponent implements OnInit {
...
   @Input()
   case: string;
+  @Input()
+  hasPrint = true;
+  @Input()
+  hasCaseHistory = true;
+  @Input()
+  hasEventSelector = true;
 
+  isPrintPartOfCaseView = false;
```
andrewfolga: make sure these do not need cleanup

veeresha-hmcts:
Verified locally, ready to merge in.

PR #285, opened 2019-03-27
Rdm 4131 new contract

adnanakgun:

https://tools.hmcts.net/jira/browse/RDM-4131

### Change description ###

Removed formgroup as part of the new contract

```
[ ] Yes
[x] No
```


src/shared/components/search-result/search-result.component.ts
```diff
@@ -29,9 +28,6 @@ export class SearchResultComponent implements OnChanges {
   @Input()
   caseState: CaseState;
 
-  @Input()
-  caseFilterFG: FormGroup;
```
andrewfolga: This has to come back.
See - RDM-1560 Fix for Work basket filters not applied for subsequent pages

Both Search and Workbasket filters are required to pass that down.

src/shared/components/search-result/search-result.component.ts
```diff
@@ -132,12 +127,10 @@ export class SearchResultComponent implements OnChanges {
   }
 
   goToPage(page): void {
-    this.hideRows = true;
```
andrewfolga: Why are we removing this?

PR #290, opened 2019-04-02
RDM-3525 - disable submit button once submitted on a case event

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-3525
https://tools.hmcts.net/jira/browse/RDM-4486
**INC0443968**

### Change description ###
Disable submit button for case event after it is clicked 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-submit/case-edit-submit.html
+++ b/src/shared/components/case-editor/case-edit-submit/case-edit-submit.html
@@ -90,8 +90,8 @@
     </fieldset>
   </ng-container>
   <div class="form-group form-group-related">
-    <button class="button button-secondary" type="button" [disabled]="!hasPrevious()" (click)="previous()">Previous</button>
+    <button class="button button-secondary" type="button" [disabled]="!hasPrevious() || isSubmitting" (click)="previous()">Previous</button>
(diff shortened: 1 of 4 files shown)


andrewfolga:
LGTM.

As confirmed with @gbenadikar the other part of the solution may come that will deal with a potential issue of multiple retries causing multiple payments.

src/shared/components/case-editor/case-edit-submit/case-edit-submit.html
```diff
@@ -91,7 +91,7 @@ <h2 class="heading-h2">Check your answers</h2>
   </ng-container>
   <div class="form-group form-group-related">
     <button class="button button-secondary" type="button" [disabled]="!hasPrevious()" (click)="previous()">Previous</button>
-    <button type="submit" [disabled]="isDisabled()" class="button">{{triggerText}}</button>
+    <button type="submit" [disabled]="isDisabled() || isSubmitting" class="button">{{triggerText}}</button>
```
mario-paniccia: we could perhaps introduce this check inside that isDisabled() method itself?

gbenadikar: actually good point, refactored

PR #297, opened 2019-04-05
RDM-3611 - register readonly fields with the form group

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-3611

### Change description ###
Persist readonly case fields on events

```
[ ] Yes
[x] No
```


src/shared/components/palette/base-field/field-write.component.ts
```diff
@@ -23,30 +22,19 @@ import { FormValidatorsService } from '../../../services/form/form-validators.se
 })
 export class FieldWriteComponent extends AbstractFieldWriteComponent implements OnInit {
 
-  @Input()
-  formGroup: FormGroup;
-
   @ViewChild('fieldContainer', {read: ViewContainerRef})
   fieldContainer: ViewContainerRef;
 
-  private defaultControlRegistrer(formGroup: FormGroup,
```
gbenadikar: moved this method to AbstractFormFieldComponent

src/shared/components/palette/base-field/field-write.component.ts
```diff
@@ -23,30 +22,19 @@ import { FormValidatorsService } from '../../../services/form/form-validators.se
...
-      formGroup.addControl(caseField.id, control);
-      return control;
-    };
-  }
-
   constructor(private resolver: ComponentFactoryResolver,
               private paletteService: PaletteService,
               private formValidatorsService: FormValidatorsService) {
     super();
   }
```
gbenadikar: add validators to write fields

src/shared/components/palette/base-field/abstract-form-field.component.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  @Input()
+  registerControl?: <T extends AbstractControl> (control: T) => T;
+
+  protected defaultControlRegister(): (control: FormControl) => AbstractControl {
+    return control => {
+      if (!this.formGroup) {
+        return null;
+      }
+      if (this.formGroup.controls[this.caseField.id]) {
```
andrewfolga: Will the `this.caseField` always be there?

src/shared/domain/definition/case-field.model.ts
```diff
@@ -26,4 +26,5 @@ export class CaseField implements Orderable {
 
   @Type(() => WizardPageField)
   wizardProps?: WizardPageField;
+  metadata?: boolean;
```
andrewfolga: Why this as part of this story?

src/shared/components/palette/complex/read-complex-field.component.ts
```diff
@@ -11,6 +11,7 @@ export class ReadComplexFieldComponent extends AbstractFieldReadComponent implem
   public paletteContext = PaletteContext;
 
   ngOnInit(): void {
+    super.ngOnInit();
```
andrewfolga: Do we need this for other field types too? I can only see this for 3 types now.

andrewfolga:
Do we need the same done for collection type as well?

src/shared/components/palette/base-field/abstract-form-field.component.ts
gbenadikar: I abstracted out this class from AbstractFieldWriteComponent and AbstractFieldReadComponent, both had casefield mandatory

src/shared/components/palette/complex/read-complex-field.component.ts
gbenadikar: Not all read components have ngOnit, whichever I have left were not being used in edit context.

src/shared/domain/definition/case-field.model.ts
gbenadikar: I needed to skip metadata fields from being registered to formGroup. Metadata fields are actually coming through as part of caseFields

src/shared/components/palette/complex/read-complex-field.component.ts
andrewfolga: Hmm not sure I understand the _whichever I have left were not being used in edit context._ Maybe I 
 am misunderstanding the requirements

gbenadikar: Components like ReadEmailFieldComponent didn't implement OnInit. Since it is now inheriting AbstractFieldReadComponent which implements OnInit, I didn't explicitly add
`ngOnInit(): void {
    super.ngOnInit();
}`
to such components. Unless it is good practice to explicitly implement OnInit for components (rather than inheriting), I won't add it. Please let me know

gbenadikar:
> Do we need the same done for collection type as well?
Read collection is not used on case edit context, so skipped it. For some reason write collection is used for all contexts as per case-edit-form.html. Let me know if you think read collection is being used for edit context.

src/shared/components/palette/base-field/abstract-form-field.component.ts
```diff
@@ -0,0 +1,34 @@
...
+import { CaseField } from '../../../domain/definition';
+import { AbstractControl, FormControl, FormGroup } from '@angular/forms';
+
+export abstract class AbstractFormFieldComponent {
+
+  @Input()
+  caseField: CaseField;
+
+  @Input()
+  formGroup?: FormGroup;
```
mario-paniccia: is there any case where for a form field `formGroup` is not set?

src/shared/components/palette/base-field/abstract-form-field.component.ts
```diff
@@ -0,0 +1,34 @@
...
+        return this.formGroup.get(this.caseField.id);
+      }
+      this.addValidators(this.caseField, control);
+      this.formGroup.addControl(this.caseField.id, control);
+      return control;
+    };
+  }
+
+  protected addValidators(caseField: CaseField, control: FormControl): void {
+    // No validators by default, override this method to add validators to the form control
```
mario-paniccia: only write field require validation am I correct? In that case I was thinking whether we could remove the addValidators logic from here. And leave it only in the write field subclass. Also removing 
`this.addValidators(this.caseField, control);` 
from the default control register. The write field subclass could create a custom control register that calls the default and then register validators to the validator service

src/shared/components/palette/base-field/abstract-form-field.component.ts
gbenadikar: FieldReadComponent is used on case view, search results etc. where form-group and registerControl do not apply, hence optional.

src/shared/components/palette/base-field/abstract-form-field.component.ts
gbenadikar: defaultControlRegister returns AbstractControl , so just wanted to avoid downcasting AbstractControl to FormControl in field write component as FormValidatorsService expects FormControl. Do you still want me to do that or leave it as it is?

andrewfolga:
LGTM

PR #316, opened 2019-04-16
Rdm 4574 definitions service

adnanakgun:

https://tools.hmcts.net/jira/browse/RDM-4574

### Change description ###

Extracted definitionsService from man-web and implemented it here. Modified createCaseFilters to consume the service.

```
[x] Yes
[ ] No
```


src/shared/components/create-case-filters/create-case-filters.component.ts
```diff
@@ -40,22 +40,20 @@ export class CreateCaseFiltersComponent implements OnChanges {
...
     this.selected = {};
-    this.selectedJurisdictionCaseTypes = [];
-    this.selectedCaseTypeEvents = [];
     this.initControls();
-  }
-
-  ngOnChanges(changes?: SimpleChanges): void {
-    if (changes.jurisdictions && changes.jurisdictions.currentValue) {
-      if (this.jurisdictions.length > 0 && this.filterJurisdictionControl) {
+    this.definitionsService.getJurisdictions(CREATE_ACCESS)
```
andrewfolga: Can we make sure we unsubscribe in OnDestroy event cycle? Obtain the subscription to a instance variable and call unsubscribe in ngOnDestroy.

PR #317, opened 2019-04-16
RDM-3325/4591: Fixed to allow the special characters for validations

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-4591

### Change description ###
Fixed to allow the special character for validations

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/services/form/form-validators.service.spec.ts
+++ b/src/shared/services/form/form-validators.service.spec.ts
@@ -43,4 +43,50 @@ describe('FormValidatorsService', () => {
     result.updateValueAndValidity();
     expect(result.invalid).toBeTruthy();
   });
+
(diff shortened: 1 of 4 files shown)


src/shared/services/form/form-validators.service.spec.ts
```diff
@@ -16,6 +16,16 @@ describe('FormValidatorsService', () => {
     expect(result.invalid).toBeFalsy();
   });
```
mario-paniccia: please add the following test: 
field mandatory, BUT, with a regex that allows all whitespaces 
what happens in that case?
we should have a validation error, because the field is mandatory. Even if the regex in theory allows it

PR #322, opened 2019-04-23
Rdm 4535 new contract

adnanakgun:

https://tools.hmcts.net/jira/browse/RDM-4535
https://tools.hmcts.net/jira/browse/RDM-4576

### Change description ###

Removed router actions from component and introduced DefinitionService

```
[x] Yes
[ ] No
```


src/shared/components/workbasket-filters/workbasket-filters.component.ts
```diff
@@ -0,0 +1,239 @@
...
+    private alertService: AlertService,
+    private windowService: WindowService,
+    private definitionsService: DefinitionsService,
+  ) {
+  }
+
+  ngOnInit(): void {
+    this.selected = {};
+
+    this.definitionsService.getJurisdictions(READ_ACCESS)
```
andrewfolga: Remember man web has to use the old component due to performance reasons. Are we extracting the old component as well having two effectively or we moved this one and changed?

src/shared/components/workbasket-filters/workbasket-filters.component.html
```diff
@@ -0,0 +1,39 @@
+<h2 class="heading-h2" aria-label="Filters">Filters</h2>
```
andrewfolga: Should we not have a conditional logic to check the jursidcitions are loaded `isDataLoaded()` before the template can be displayed? This would be like we have on the other extracted components that pull data inside e.g. case-create.component.html or case-history.component.html to name a few.

src/shared/components/workbasket-filters/workbasket-filters.component.ts
adnanakgun: Are we going to have two components then? Or will the component decide whether to use the service?

andrewfolga: I think two components is better as we do not need to regression test man web that way so will save us time. It would make the extraction consistent with all other ones this way. It would just be a new wrapper component then.

adnanakgun: Cool, I'll just rename this component then.

src/shared/components/workbasket-filters/workbasket-filters.component.html
adnanakgun: Done

src/shared/components/workbasket-filters/workbasket-filters.component.ts
adnanakgun: Done

src/shared/components/case-list-filters/case-list-filters.component.html
```diff
@@ -0,0 +1,41 @@
+<ng-container *ngIf="jurisdictions">
```
andrewfolga: This is supposed to wrap the workbasket filters component similar to search filter having a wrapper component. Not sure why we have a copy here?

src/shared/components/case-list-filters/case-list-filters.component.ts
```diff
@@ -0,0 +1,241 @@
...
+  DefinitionsService } from '../../services';
+
+const FORM_GROUP_VAL_LOC_STORAGE = 'case-list-filters-form-group-value';
+const SAVED_QUERY_PARAM_LOC_STORAGE = 'savedQueryParams';
+@Component({
+  selector: 'ccd-case-list-filters',
+  templateUrl: './case-list-filters.component.html',
+  styleUrls: ['./case-list-filters.component.scss']
+})
+export class CaseListFiltersComponent implements OnInit {
```
andrewfolga: The logic here should be much simpler without duplicating the workbasket filters component but just wrapping it similar to search filters and search filters wrapper component.

src/shared/components/case-list-filters/case-list-filters.component.html
adnanakgun: Done.

src/shared/components/case-list-filters/case-list-filters.component.ts
adnanakgun: Done

adnanakgun:
Please re-review @andrewfolga

src/shared/components/case-list-filters/case-list-filters.component.spec.ts
```diff
@@ -1,1048 +0,0 @@
-import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
andrewfolga: For completeness I think we do need unit tests checking emitting apply and reset notifications and rendering of wrapped component once data is loaded?

adnanakgun: Just added tests

adnanakgun:
Please re-review @andrewfolga

PR #324, opened 2019-04-25
RDM-4286-Not Equals and RDM-2804-OR logic

ganganna-peddapothu:
### https://tools.hmcts.net/jira/browse/RDM-4286 & https://tools.hmcts.net/jira/browse/RDM-2804  ###

```
[ ] Yes
[ X] No
```


ganganna-peddapothu:
This PR is replacement of https://github.com/hmcts/ccd-case-ui-toolkit/pull/320 which was reviewed.

To resolve conflicts I have replaced this PR.

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -4,21 +4,34 @@ import { FieldsUtils } from '../../../services/fields';
...
+  private static CONDITION_EQUALS = '=';
   private static readonly CONTAINS = 'CONTAINS';
 
   static addPathPrefixToCondition(showCondition: string, pathPrefix): string {
     if (!pathPrefix || pathPrefix === '') {
       return showCondition;
     }
-    let andConditions = showCondition.split(ShowCondition.AND_CONDITION_REGEXP);
-    andConditions = andConditions.map(condition => {
+    if (showCondition.search(ShowCondition.OR_CONDITION_REGEXP) !== -1) {
```
rafalkalita: What will happen when we will mix OR and AND condition? It will only add prefix to OR? Please update test with this scenario.

ganganna-peddapothu: There is no business of mixing OR and AND operations in the same condition. Steve confirmed.

src/shared/directives/conditional-show/domain/conditional-show.model.spec.ts
```diff
@@ -607,4 +607,102 @@ describe('conditional-show', () => {
...
+    });
+    it('Scenario2 show: is not blank', () => {
+      let sc = new ShowCondition('field!=""');
+      let fields = {
+        field: 'MOJ'
+      };
+      let matched = sc.match(fields);
+      expect(matched).toBe(true);
+    });
+    it('Scenario2 hide: is not blank', () => {
```
mario-paniccia: can we have a test 'is not blank multiple spaces' i.e.
`'field!="     "'`

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -33,18 +46,27 @@ export class ShowCondition {
...
+    } else {
+      let andConditions = condition.split(ShowCondition.AND_CONDITION_REGEXP);
+      return andConditions.every(andCondition => this.matchEqualityCondition(fields, andCondition, path));
+    }
   }
 
   private matchEqualityCondition(fields: any, condition: string, path?: string): boolean {
     if (condition.search(ShowCondition.CONTAINS) === -1) {
-      let field = condition.split('=')[0];
+      let conditionSeparaor = ShowCondition.CONDITION_EQUALS;
```
mario-paniccia: typo here `conditionSeparaor`

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -4,21 +4,34 @@ import { FieldsUtils } from '../../../services/fields';
 export class ShowCondition {
```
mario-paniccia: the logic seems ok to me. But this class has grown a lot in complexity over time. I can see that you did some good restructuring, but I think it's still a bit complex. Any way we can further simplify it @ganganna-peddapothu? for example introduce more private methods with descriptive names, or brake it down in multiple classes, etc...

src/shared/directives/conditional-show/domain/conditional-show.model.spec.ts
ganganna-peddapothu: DONE

PR #326, opened 2019-04-26
Rdm 4537 doc demo

adnanakgun:

https://tools.hmcts.net/jira/browse/RDM-4537

### Change description ###

Added demos for workbasket and case list filters

```
[ ] Yes
[x] No
```


demo/src/app/workbasket-filters-consumer.component.ts
```diff
@@ -0,0 +1,120 @@
+import { Component, OnInit } from '@angular/core';
```
andrewfolga: The only extracted component we expose should be the new case list filters. I think we should remove workbasket filters from the demo app as discussed before.

src/shared/components/case-list-filters/case-list-filters.component.html
```diff
@@ -0,0 +1,41 @@
+<ng-container *ngIf="jurisdictions">
```
andrewfolga: Case list filters should be reusing workbasket filters not copying it similar to search filters and search filters wrapper.

demo/src/app/workbasket-filters-consumer.component.ts
adnanakgun: Done

src/shared/components/case-list-filters/case-list-filters.component.html
adnanakgun: Done

adnanakgun:
Please re-review @andrewfolga

src/shared/components/case-list-filters/case-list-filters.component.ts
```diff
@@ -0,0 +1,47 @@
+import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { READ_ACCESS } from '../../domain/case-view/access-types.model';
+import { DefinitionsService } from '../../services';
+import { Jurisdiction } from '../../domain';
+
+@Component({
+  selector: 'ccd-case-list-filters',
+  templateUrl: './case-list-filters.component.html'
+})
+export class CaseListFiltersComponent implements OnInit {
```
andrewfolga: Unit tests checking emitting apply and reset notifications?

adnanakgun: added tests

adnanakgun:
Please re-review @andrewfolga

PR #327, opened 2019-04-28
Rdm 4498 - Add new DynamicList base type

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4498

As a user
I am able to see a 'DynamicList' populated with data dynamically
So that the options that are presented in the list are move relevant to the action I am doing
    User is able to populate a DynamicList on the AboutToStart callback
    User is able to populate a DynamicList on the MidEvent callback
    User is able to populate a DynamicList on the AboutToSubmit callback
    If in the callback the data exceeds 250 character length then an error message is returned on the frontend
    If in the callback the data for the key exceeds the 150 then an error message is returned on the frontend
    Callback response has no data to populate list - we display an empty list container label 'Empty' - mandatory is kept (user is not able to progress because no selection can be made)
    If no value is selected from the list then we save it as 'null'
    User is able to use DynamicList in
        Collection
        ComplexType
        Nested Collection/ComplexTypes
        Anywhere where fixed list can be used currently
    User is able to see selected option from Dynamic List in case details
    Selected option and other options are saved in the DB
    Show and Hide will work as is
    Event History needs reflect the selected value from the list

```
[ ] Yes
[X] No
```


src/shared/components/palette/dynamic-list/read-dynamic-list-field.component.ts
```diff
@@ -0,0 +1,11 @@
+import { AbstractFieldReadComponent } from '../base-field/abstract-field-read.component';
+import { Component } from '@angular/core';
+
+@Component({
+  selector: 'ccd-read-dynamic-list-field',
+  template: '<span *ngIf="caseField.value" class="text-16">' +
+    '{{caseField.value.default.code | ccdDynamicList:caseField.value.dynamic_list_items}}' +
+    '</span>',
+})
+export class ReadDynamicListFieldComponent extends AbstractFieldReadComponent {
```
mario-paniccia: do we really need to introduce new palette components for Dynamic List?
Can't we reuse FixedList components? 
In both cases we just have to display a dropdown so what we render on the page is exactly the same

src/shared/components/palette/fixed-list/read-fixed-list-field.component.ts
```diff
@@ -3,6 +3,23 @@ import { Component } from '@angular/core';
...
 @Component({
   selector: 'ccd-read-fixed-list-field',
-  template: '<span class="text-16">{{caseField.value | ccdFixedList:caseField.field_type.fixed_list_items}}</span>',
+  template: '<span class="text-16">{{getValue() | ccdFixedList: getLists()}}</span>',
 })
-export class ReadFixedListFieldComponent extends AbstractFieldReadComponent {}
+export class ReadFixedListFieldComponent extends AbstractFieldReadComponent {
+
+  getValue() {
+    if (this.caseField.field_type.type === 'DynamicList') {
```
mario-paniccia: please move this whole block of code inside `CaseField` class in an accessor method. Have a look at https://www.typescriptlang.org/docs/handbook/classes.html section 'Accessors'. Let's add a 
```
private _value: string;

    get value(): string {
        if (this.caseField.field_type.type === 'DynamicList') {
     return this._value ? this._value.value.code : '';
    } else {
      return this._value;
    }
    }
```

src/shared/components/palette/fixed-list/read-fixed-list-field.component.ts
```diff
@@ -3,6 +3,23 @@ import { Component } from '@angular/core';
...
+      return this.caseField.value ? this.caseField.value.value.code : '';
+    } else {
+      return this.caseField.value;
+    }
+  }
+
+  getLists() {
```
mario-paniccia: same as above please, let's move this in an accessor method in `CaseField`

src/shared/components/palette/palette.service.ts
```diff
@@ -61,6 +61,7 @@ export class PaletteService {
       case 'MoneyGBP':
         return write ? WriteMoneyGbpFieldComponent : ReadMoneyGbpFieldComponent;
       case 'FixedList':
+      case 'DynamicList':
```
mario-paniccia: 👍

src/shared/domain/definition/field-type.model.ts
```diff
@@ -1,4 +1,5 @@
 import { FixedListItem } from './fixed-list-item.model';
+import { DynamicListItem } from './dynamic-list-item.model';
```
mario-paniccia: is this still used?

src/shared/domain/definition/dynamic-list-item.model.ts
```diff
@@ -0,0 +1,4 @@
+export class DynamicListItem {
```
mario-paniccia: is this still used?

src/shared/components/palette/fixed-list/read-fixed-list-field.component.ts
mario-paniccia: Then please move the test code in read-fixed-list-field.component.spec.ts  to the one for CaseField

src/shared/services/form/form-value.service.ts
```diff
@@ -69,6 +69,12 @@ export class FormValueService {
     Object.keys(cloneForm['data']).forEach((key) => {
       if (caseFields.findIndex((element) => element.id === key) < 0) {
         delete cloneForm['data'][key];
+      } else {
+        caseFields.forEach(function (e) {
+          if (e.id === key && e.field_type && e.field_type.type === 'DynamicList') {
```
mario-paniccia: I don't think `e.field_type` null check is required because field always has  a field_type

src/shared/services/form/form-value.service.ts
```diff
@@ -69,6 +69,12 @@ export class FormValueService {
     Object.keys(cloneForm['data']).forEach((key) => {
       if (caseFields.findIndex((element) => element.id === key) < 0) {
         delete cloneForm['data'][key];
+      } else {
```
mario-paniccia: why is this necessary can you explain please?

src/shared/domain/definition/field-type.model.ts
mohammedanas: removed

src/shared/components/case-editor/case-edit-page/case-edit-page.component.spec.ts
```diff
@@ -522,6 +522,7 @@ describe('CaseEditPageComponent', () => {
       spyOn(caseEditComponentStub, 'form');
       spyOn(caseEditComponentStub, 'validate').and.returnValue(of(validateResult));
       spyOn(formValueService, 'sanitise').and.returnValue(eventData);
+      spyOn(formValueService, 'sanitiseDynamicLists').and.returnValue(eventData);
```
mario-paniccia: can we add a verifycation that this method is called?

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -123,6 +123,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       this.isSubmitting = true;
       this.error = null;
       let currentPageFields = this.formValueService.filterCurrentPageFields(this.currentPage.case_fields, this.editForm.value);
```
mario-paniccia: can you please rename `currentPageFields` to `pageFormFields`

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -123,6 +123,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
...
+      this.formValueService.sanitiseDynamicLists(this.currentPage.case_fields,currentPageFields);
```
mario-paniccia: missing space between the 2 arguments

src/shared/services/form/form-value.service.ts
```diff
@@ -74,4 +74,26 @@ export class FormValueService {
     return cloneForm;
   }
 
+  sanitiseDynamicLists(caseFields: CaseField[], editForm: any): any {
+    // loop thorught dynamic fields in caseFields
```
mario-paniccia: This code is quite complex. I was thinking we should move it into a class called `ListFieldTypeSanitiser`. And `sanitiseDynamicLists` in `form-value.service.ts` will simply call `listFieldTypeSanitiser.sanitiseLists`. When moving the code in that class, you should remove the comments and replace them with private methods with self explanatory names. Rather than having all in a long method

We can do this as part of the tech debt ticket. Will update the description

src/shared/components/palette/dynamic-list/read-dynamic-list-field.component.ts
mohammedanas: removed

src/shared/components/palette/fixed-list/read-fixed-list-field.component.ts
mohammedanas: Applied the code as per discussion

src/shared/components/palette/fixed-list/read-fixed-list-field.component.ts
mohammedanas: Will address accessor method in fixed-list

src/shared/domain/definition/dynamic-list-item.model.ts
mohammedanas: removed

src/shared/services/form/form-value.service.ts
mohammedanas: removed

src/shared/services/form/form-value.service.ts
mohammedanas: removed

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
mohammedanas: applied changes

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
mohammedanas: reformatted code

src/shared/services/form/form-value.service.ts
mohammedanas: oh ok thanks

src/shared/components/case-editor/case-edit-page/case-edit-page.component.spec.ts
mohammedanas: called now

PR #347, opened 2019-05-16
Rdm 4704 - Add accessor methods in CaseField and related changes across application

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4704

Add accessor methods in CaseField and related changes across application.
Ie : Fix compatiblity issues across application

CaseField in the toolkit has an optional 'value' property.
We need to implement the optional 'value' as a mandatory property with accessor methods. And move the implementation of getValue() method into the getter method for value. i.e.
_value: any;
get value() { move here current content of getValue() method }
set value() {
getValue() { .... } remove this
This is needed because getValue() currently has logic to fetch the value from dynamic lists. And we want this logic to be used everywhere we currently use field.value
Doing this will break our test code in multiple places so we are tackling this in this ticket. 
Other things to do:
1) remove method instantiate() in  write-fixed-list-field.component.ts 
2) introduce a 'FieldTypeSanitiser' as per comment in https://github.com/hmcts/ccd-case-ui-toolkit/pull/327/files

```
[X] Yes
[ ] No
```

### DIFF
--- a/src/shared/components/case-viewer/case-view/case-view.component.ts
+++ b/src/shared/components/case-viewer/case-view/case-view.component.ts
@@ -5,6 +5,7 @@ import { CasesService, CaseService } from '../../case-editor';
 import { DraftService } from '../../../services';
 import { Observable, throwError } from 'rxjs';
 import { map } from 'rxjs/operators';
+import { plainToClass } from 'class-transformer';
 
 @Component({
   selector: 'ccd-case-view',
--- a/src/shared/components/case-viewer/case-view/case-view.component.ts
+++ b/src/shared/components/case-viewer/case-view/case-view.component.ts
@@ -1,11 +1,11 @@
-import { Component, Input, OnInit, Output, EventEmitter } from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
 import { AlertService } from '../../../services/alert';
 import { CaseView, Draft } from '../../../domain';
-import { CasesService, CaseService } from '../../case-editor';
+import { CaseService, CasesService } from '../../case-editor';
 import { DraftService } from '../../../services';
 import { Observable, throwError } from 'rxjs';
 import { map } from 'rxjs/operators';
-import { plainToClass } from 'class-transformer';
+import { plainToClassFromExist } from 'class-transformer';
 
 @Component({
   selector: 'ccd-case-view',
@@ -33,8 +33,8 @@ export class CaseViewComponent implements OnInit {
     this.getCaseView(this.case)
       .pipe(
         map(caseView => {
-          this.caseDetails = caseView;
-          this.caseService.announceCase(caseView);
+          this.caseDetails = plainToClassFromExist(new CaseView(), caseView);
+          this.caseService.announceCase(this.caseDetails);
         })
       )
       .toPromise()
--- a/src/shared/components/case-viewer/case-viewer.component.spec.ts
+++ b/src/shared/components/case-viewer/case-viewer.component.spec.ts
@@ -16,15 +16,15 @@ import { LabelSubstitutorDirective } from '../../directives/substitutor';
 import { HttpError } from '../../domain/http';
 import { OrderService } from '../../services/order';
 import { DeleteOrCancelDialogComponent } from '../../components/dialogs';
-import { CaseViewTrigger, CaseViewEvent, CaseView } from '../../domain/case-view';
+import { CaseView, CaseViewEvent, CaseViewTrigger } from '../../domain/case-view';
 import { AlertService } from '../../services/alert';
 import { CallbackErrorsContext } from '../../components/error/domain';
 import { DraftService } from '../../services/draft';
 import { CaseReferencePipe } from '../../pipes/case-reference';
+import { MatDialog, MatDialogConfig, MatDialogRef } from '@angular/material';
+import { CaseService } from '../case-editor';
 import createSpyObj = jasmine.createSpyObj;
 import any = jasmine.any;
-import { MatDialog, MatDialogRef, MatDialogConfig } from '@angular/material';
-import { CaseService } from '../case-editor';
 
 @Component({
   // tslint:disable-next-line
@@ -132,7 +132,7 @@ const EVENTS: CaseViewEvent[] = [
 ];
 
 const METADATA: CaseField[] = [
-  {
+  Object.assign(new CaseField(), {
     id: '[CASE_REFERENCE]',
     label: 'Case Reference',
     value: 1533032330714079,
@@ -153,8 +153,8 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  },
-  {
+  }),
+  Object.assign(new CaseField(), {
     id: '[CASE_TYPE]',
     label: 'Case Type',
     value: 'DIVORCE',
@@ -175,8 +175,8 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  },
-  {
+  }),
+  Object.assign(new CaseField(), {
     id: '[CREATED_DATE]',
     label: 'Created Date',
     value: '2018-07-31T10:18:50.737',
@@ -197,8 +197,8 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  },
-  {
+  }),
+  Object.assign(new CaseField(), {
     id: '[JURISDICTION]',
     label: 'Jurisdiction',
     value: 'DIVORCE',
@@ -219,8 +219,8 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  },
-  {
+  }),
+  Object.assign(new CaseField(), {
     id: '[LAST_MODIFIED_DATE]',
     label: 'Last Modified Date',
     value: '2018-07-31T10:18:50.737',
@@ -241,8 +241,8 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  },
-  {
+  }),
+  Object.assign(new CaseField(), {
     id: '[SECURITY_CLASSIFICATION]',
     label: 'Security Classification',
     value: 'PUBLIC',
@@ -263,8 +263,8 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  },
-  {
+  }),
+  Object.assign(new CaseField(), {
     id: '[SECURITY_CLASSIFICATION]',
     label: 'Security Classification',
     value: 'PUBLIC',
@@ -285,7 +285,7 @@ const METADATA: CaseField[] = [
     show_condition: null,
     show_summary_change_option: null,
     show_summary_content_option: null
-  }
+  })
 ];
 
 const TRIGGERS: CaseViewTrigger[] = [
@@ -340,7 +340,7 @@ const CASE_VIEW: CaseView = {
       label: 'Name',
       order: 2,
       fields: [
-        {
+        Object.assign(new CaseField(), {
           id: 'PersonFirstName',
           label: 'First name',
           display_context: 'OPTIONAL',
@@ -352,8 +352,8 @@ const CASE_VIEW: CaseView = {
           value: 'Janet',
           show_condition: '',
           hint_text: ''
-        },
-        {
+        }),
+        Object.assign(new CaseField(), {
           id: 'PersonLastName',
           label: 'Last name',
           display_context: 'OPTIONAL',
@@ -365,8 +365,8 @@ const CASE_VIEW: CaseView = {
           value: 'Parker',
           show_condition: 'PersonFirstName="Jane*"',
           hint_text: ''
-        },
-        {
+        }),
+        Object.assign(new CaseField(), {
           id: 'PersonComplex',
           label: 'Complex field',
           display_context: 'OPTIONAL',
@@ -378,7 +378,7 @@ const CASE_VIEW: CaseView = {
           order: 3,
           show_condition: 'PersonFirstName="Park"',
           hint_text: ''
-        }
+        })
       ],
       show_condition: 'PersonFirstName="Janet"'
     },
@@ -386,7 +386,7 @@ const CASE_VIEW: CaseView = {
       id: 'HistoryTab',
       label: 'History',
       order: 1,
-      fields: [{
+      fields: [Object.assign(new CaseField(), {
         id: 'CaseHistory',
         label: 'Case History',
         display_context: 'OPTIONAL',
@@ -398,7 +398,7 @@ const CASE_VIEW: CaseView = {
         value: EVENTS,
         show_condition: '',
         hint_text: ''
-      }],
+      })],
       show_condition: ''
     },
     {
--- a/src/shared/components/case-viewer/case-viewer.component.ts
+++ b/src/shared/components/case-viewer/case-viewer.component.ts
@@ -20,7 +20,6 @@ import { MatDialog, MatDialogConfig } from '@angular/material';
 import { CaseService } from '../case-editor';
 import { Type } from 'class-transformer';
 
-// @dynamic
 @Component({
   selector: 'ccd-case-viewer',
   templateUrl: './case-viewer.component.html',
@@ -39,9 +38,7 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
   BANNER = DisplayMode.BANNER;
 
   caseDetails: CaseView;
-  @Type(() => CaseTab)
   sortedTabs: CaseTab[];
-  @Type(() => CaseField)
   caseFields: CaseField[];
   error: any;
   triggerTextStart = CaseViewerComponent.TRIGGER_TEXT_START;
--- a/src/shared/components/case-viewer/case-viewer.component.ts
+++ b/src/shared/components/case-viewer/case-viewer.component.ts
@@ -18,7 +18,6 @@ import { CallbackErrorsContext } from '../../components/error/domain';
 import { DraftService } from '../../services/draft';
 import { MatDialog, MatDialogConfig } from '@angular/material';
 import { CaseService } from '../case-editor';
-import { Type } from 'class-transformer';
 
 @Component({
   selector: 'ccd-case-viewer',
@@ -93,8 +92,8 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
   private init() {
     // Clone and sort tabs array
     this.sortedTabs = this.orderService.sort(this.caseDetails.tabs);
-    let tabFields = this.getTabFields()
-    this.caseFields = this.reinstantiateCaseFields(tabFields);
+
+    this.caseFields = this.getTabFields();
 
     this.sortedTabs = this.sortTabFieldsAndFilterTabs(this.sortedTabs);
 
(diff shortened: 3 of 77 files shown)


src/shared/components/case-viewer/case-viewer.component.ts
```diff
@@ -8,19 +8,19 @@ import { Observable } from 'rxjs';
...
-import { DRAFT_QUERY_PARAM } from '../../domain';
 import { AlertService } from '../../services/alert';
 import { CallbackErrorsContext } from '../../components/error/domain';
 import { DraftService } from '../../services/draft';
 import { MatDialog, MatDialogConfig } from '@angular/material';
 import { CaseService } from '../case-editor';
-import { NgZone } from '@angular/core';
+import { Type } from 'class-transformer';
 
+// @dynamic
```
mario-paniccia: why this comment? what's the meaning?

src/shared/components/case-viewer/case-viewer.component.ts
```diff
@@ -39,7 +39,9 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
   BANNER = DisplayMode.BANNER;
 
   caseDetails: CaseView;
+  @Type(() => CaseTab)
```
mario-paniccia: is this a leftover? Usually this annotation should be used only on domain classes, not on a component

src/shared/components/case-viewer/case-viewer.component.ts
```diff
@@ -39,7 +39,9 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
...
   sortedTabs: CaseTab[];
+  @Type(() => CaseField)
```
mario-paniccia: is this a leftover? Usually this annotation should be used only on domain classes, not on a component

src/shared/components/case-viewer/case-viewer.component.ts
```diff
@@ -94,8 +96,8 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
   private init() {
     // Clone and sort tabs array
     this.sortedTabs = this.orderService.sort(this.caseDetails.tabs);
-
-    this.caseFields = this.getTabFields();
+    let tabFields = this.getTabFields()
+    this.caseFields = this.reinstantiateCaseFields(tabFields);
```
mario-paniccia: don't think manually reinstantiating the case fields is the right thing to do. We should instead look at the origin of those fields, and why they were not class instances at the origin. And fix it there.
For example could we apply the class transform library to `case-tab.model.ts`? Finding out why that model class gets deserialised and do a `plainToClass` there. In that case, all this code introduced here is no longer needed

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -48,15 +48,15 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
   }
 
   buildCaseField(item, index: number): CaseField {
-    return {
+    return Object.assign(new CaseField(),{
```
mario-paniccia: just to improve readability, can you create a private method in this class and call it `newCaseField` and then do the 
`Object.assign(new CaseField()` there?

src/shared/components/palette/complex/fields-filter.pipe.ts
```diff
@@ -82,7 +82,7 @@ export class FieldsFilterPipe implements PipeTransform {
 
     return fields
       .map(f => {
-        let clone = { ...f };
+        let clone = Object.assign({}, f);
```
mario-paniccia: if you look at field.utils.ts class there is a method `cloneObject` that does exactly this. Can you make it `public` and call that here please?

src/shared/components/palette/fixed-list/read-fixed-list-field.component.ts
```diff
@@ -1,39 +1,9 @@
 import { AbstractFieldReadComponent } from '../base-field/abstract-field-read.component';
-import { Component, OnInit } from '@angular/core';
-import { CaseField } from '../../../domain/definition';
+import { Component } from '@angular/core';
 
 @Component({
   selector: 'ccd-read-fixed-list-field',
-  template: '<span class="text-16">{{caseField.getValue() | ccdFixedList:caseField.getItems()}}</span>',
+  template: '<span class="text-16">{{caseField.value | ccdFixedList:caseField.items}}</span>',
```
mario-paniccia: I think we should rename `items` to `listItems`

src/shared/components/search-result/search-result.component.ts
```diff
@@ -146,13 +147,13 @@ export class SearchResultComponent implements OnChanges {
   }
 
   buildCaseField(col: SearchResultViewColumn, result: SearchResultViewItem): CaseField {
-    return {
+    return plainToClass(CaseField, {
```
mario-paniccia: in the methods above we used Object.assign,  can we use the same approach in all the places? Either we use Object.assign everywhere or we use plainToClass

src/shared/domain/definition/case-field.model.ts
```diff
@@ -26,20 +26,33 @@ export class CaseField implements Orderable {
   @Type(() => WizardPageField)
   wizardProps?: WizardPageField;
 
+  private _value: any;
+  private _items: any = [];
```
mario-paniccia: can we call this `list_items` ?

src/shared/domain/definition/case-field.model.ts
```diff
@@ -26,20 +26,33 @@ export class CaseField implements Orderable {
...
 
-  getValue?(): string {
-    if (this.field_type.type === 'DynamicList') {
-      return this.value ? this.value.value.code : '';
+  @Expose()
```
mario-paniccia: is @Expose() needed ?

src/shared/domain/definition/case-field.model.ts
```diff
@@ -26,20 +26,33 @@ export class CaseField implements Orderable {
...
     } else {
-      return this.value;
+      return this._value;
     }
   }
 
-  getItems?(): any {
-    if (this.field_type.type === 'DynamicList') {
-      return this.value ? this.value.list_items : [];
+  set value(newValue: any) {
```
mario-paniccia: can you rename `newValue` to `value`?

src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
```diff
@@ -15,38 +11,10 @@ export class WriteFixedListFieldComponent extends AbstractFieldWriteComponent im
...
-      cf.display_context_parameter = this.caseField.display_context_parameter;
-      cf.show_condition = this.caseField.show_condition;
-      cf.show_summary_change_option = this.caseField.show_summary_change_option;
-      cf.show_summary_content_option = this.caseField.show_summary_content_option;
-      cf.acls = this.caseField.acls;
-      cf.wizardProps = this.caseField.wizardProps;
-      this.caseField = cf;
+    this.fixedListControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null));
+    if(this.fixedListControl.value && this.fixedListControl.value.value){
+      this.fixedListControl.setValue(this.fixedListControl.value.value.code);
```
mario-paniccia: I don't understand why this is needed. The caseField.value that we call on the line above (15) returns already the `code` isn't it? 
```if (this.field_type && this.field_type.type === 'DynamicList') {
      return this._value && this._value.value ? this._value.value.code : this._value;
    } ```

File: src/shared/domain/definition/case-field.model.ts [hunk 13]
```diff
@@ -26,20 +26,33 @@ export class CaseField implements Orderable {
...
+  @Expose()
+  get items(): any {
+    if (this.field_type && this.field_type.type === 'DynamicList') {
+      return this._value.list_items ? this._value.list_items : this._items;
     } else {
       return this.field_type.fixed_list_items;
     }
   }
+
+  set items(newItems: any) {
```
mario-paniccia: can you rename `newItems` to `items` please

src/shared/domain/definition/case-field.model.ts
```diff
@@ -1,7 +1,7 @@
 import { Orderable } from '../order';
```
mario-paniccia: I think we must add some good unit tests for this class. With some examples of the json of a dynamic list

src/shared/domain/search/search-result-view-item.model.ts
```diff
@@ -1,8 +1,11 @@
 import { CaseField } from '../definition';
+import { Type } from 'class-transformer';
 
+// @dynamic
```
mario-paniccia: is this annotation needed?

src/shared/services/fields/fields.utils.ts
```diff
@@ -128,8 +128,8 @@ export class FieldsUtils {
     };
   }
 
-  public cloneObject(obj: any): any {
-    return Object.assign({}, obj);
+  public cloneObject(obj: any): CaseField {
```
mario-paniccia: the `cloneObject` method was used to clone any type of object, you changed it to clone only CaseField. 
Rather than doing that, I think we should leave `cloneObject` as it was and make it public so it can be reused everywhere. 
And you can add a new public method `cloneCaseField`

src/shared/services/fields/fields.utils.ts
```diff
@@ -141,7 +141,7 @@ export class FieldsUtils {
   }
 
   private mergeFields(caseFields: CaseField[], formFields: any, mergeFunction: (CaseField, any) => void) {
-    let result = this.cloneObject(formFields);
+    let result = Object.assign({}, formFields);
```
mario-paniccia: before was more readable

src/shared/services/form/form-value.service.ts
```diff
@@ -75,25 +76,7 @@ export class FormValueService {
...
-              {
-                value: dynamicField.value.list_items.filter(value => value.code === editForm['data'][key])[0],
-                list_items: dynamicField.value.list_items
-              };
-          }
-        });
-      });
-
-
+    return FieldTypeSanitiser.sanitiseLists(caseFields, editForm);
```
mario-paniccia: can you pass an instance of  FieldTypeSanitiser with constructor injection and call that here? i.e. `fieldTypeSanitiser.sanitiseLists` rather than `FieldTypeSanitiser.sanitiseLists`

src/shared/services/form/field-type-sanitiser.ts
```diff
@@ -0,0 +1,47 @@
+import { CaseField } from '../../domain/definition';
+import { Injectable } from '@angular/core';
+
+// @dynamic
+@Injectable()
+export class FieldTypeSanitiser {
+
+  /**
+   * This method takes finds dynamiclists in a form and replaces its string value, with
```
mario-paniccia: wrong wording here

src/shared/components/case-viewer/case-viewer.component.spec.ts
```diff
@@ -132,7 +132,7 @@ const EVENTS: CaseViewEvent[] = [
 ];
 
 const METADATA: CaseField[] = [
-  {
+  Object.assign(new CaseField(), {
```
mario-paniccia: would `<CaseField>` work here like we did above rather than doing `Object.assign`? Let's use one approach only if possible please

src/shared/components/case-viewer/case-viewer.component.ts
mohammedanas: Its to get away from errors during compilation around lambda expression

src/shared/components/palette/collection/write-collection-field.component.ts
mohammedanas: ok

src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
mohammedanas: When we choose a dynamic list and press contiue, we sanitisedynamiclists and add that to the formcontrol, for validate and going to next page now if we move previous or go back to the page from change link of check your answers it does not get selected as the formvalue contains enriched data ie { value, list_items} instead of 'List1'

src/shared/services/fields/fields.utils.ts
mohammedanas: This method was only used for case fields and locally, I felt it to be changed for it.
I will change accordingly

src/shared/services/fields/fields.utils.ts
mohammedanas: ok

src/shared/services/fields/fields.utils.ts
mohammedanas: changed

src/shared/services/fields/fields.utils.ts
mohammedanas: changed back

src/shared/domain/search/search-result-view-item.model.ts
mohammedanas: Yes

mohammedanas: Else you get lambda expression compile error when using class transformer

src/shared/domain/definition/case-field.model.ts
mohammedanas: Its needed

src/shared/domain/definition/case-field.model.ts
mohammedanas: done

src/shared/domain/definition/case-field.model.ts
mohammedanas: renamed

src/shared/domain/definition/case-field.model.ts
mohammedanas: done

src/shared/services/form/form-value.service.ts
mohammedanas: done

src/shared/components/search-result/search-result.component.ts
mohammedanas: done

src/shared/domain/definition/case-field.model.ts
mohammedanas: added

src/shared/services/form/field-type-sanitiser.ts
mohammedanas: corrected

src/shared/components/case-viewer/case-viewer.component.ts
```diff
@@ -112,9 +110,13 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
     }
   }
 
+  private reinstantiateCaseFields(caseFields: CaseField[]): CaseField[] {
```
mario-paniccia: is this a leftover?

src/shared/components/case-viewer/services/case.resolver.ts
```diff
@@ -61,7 +62,7 @@ export class CaseResolver implements Resolve<CaseView> {
         .getCaseViewV2(cid)
         .pipe(
           map(caseView => {
-            this.cachedCaseView = caseView;
+            this.cachedCaseView = plainToClassFromExist(new CaseView(), caseView);
```
mario-paniccia: wouldn't just `plainToClass` work?

src/shared/components/case-viewer/services/case.resolver.ts
```diff
@@ -75,7 +76,7 @@ export class CaseResolver implements Resolve<CaseView> {
       .getDraft(cid)
       .pipe(
         map(caseView => {
-          this.cachedCaseView = caseView;
+          this.cachedCaseView = plainToClassFromExist(new CaseView(), caseView);
```
mario-paniccia: wouldn't just plainToClass work?

src/shared/components/case-viewer/case-view/case-view.component.ts
```diff
@@ -32,8 +33,8 @@ export class CaseViewComponent implements OnInit {
     this.getCaseView(this.case)
       .pipe(
         map(caseView => {
-          this.caseDetails = caseView;
-          this.caseService.announceCase(caseView);
+          this.caseDetails = plainToClassFromExist(new CaseView(), caseView);
```
mario-paniccia: wouldn't just plainToClass work?

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -48,15 +48,19 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
   }
 
   buildCaseField(item, index: number): CaseField {
-    return {
+    return this.newCaseField(index, item);
+  }
+
+  private newCaseField(index: number, item) {
+    return Object.assign(new CaseField(), {
```
mario-paniccia: just came to my mind rather than Object.assign why not simply: 
```cf = new CaseField();
cf.id = index.ToString();
cf.field_type =  this.caseField.field_type.collection_field_type;
...

```

src/shared/components/case-viewer/case-viewer.component.ts
mohammedanas: removed and moved code to calling urls

src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
```diff
@@ -15,38 +11,14 @@ export class WriteFixedListFieldComponent extends AbstractFieldWriteComponent im
   fixedListControl: FormControl;
 
   ngOnInit() {
-    this.instantiate();
     let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
-    this.fixedListControl = this.registerControl(new FormControl(notEmpty ? this.caseField.getValue() : null));
-
+    this.fixedListControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null));
+    if (this.hasSanitisedDynamicListData()) {
```
mario-paniccia: did we find out why this is needed?

src/shared/domain/case-view/case-view.model.ts
```diff
@@ -2,7 +2,9 @@ import { CaseTab } from './case-tab.model';
 import { CaseViewEvent } from './case-view-event.model';
 import { CaseViewTrigger } from './case-view-trigger.model';
 import { CaseField } from '../definition';
+import { Type } from 'class-transformer';
 
+// @dynamic
```
mario-paniccia: is this comment required?

src/shared/components/case-viewer/services/case.resolver.ts
mohammedanas: No Plain class does not work

src/shared/components/case-viewer/services/case.resolver.ts
mohammedanas: No it didnt

src/shared/components/case-viewer/case-view/case-view.component.ts
mohammedanas: no it didnt

src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
mohammedanas: For some reason its not erasing the history

src/shared/components/case-viewer/case-viewer.component.ts
mohammedanas: removed

src/shared/components/case-viewer/case-viewer.component.spec.ts
mohammedanas: No it didnt work in this case

src/shared/components/palette/collection/write-collection-field.component.ts
mohammedanas: Depends on how strongly you feel about it,

mario-paniccia: ok let's leave it

src/shared/domain/definition/case-field.model.ts
```diff
@@ -26,20 +26,39 @@ export class CaseField implements Orderable {
...
     }
   }
+
+  set list_items(items: any) {
+    this._list_items = items;
+  }
+
+  @Expose()
+  public isReadonly() {
+    return  (!(typeof this.display_context === 'undefined' || this.display_context === null )
```
mario-paniccia: can't this `!(typeof this.display_context === 'undefined' || this.display_context === null )` be replaced simply by:
`this.display_context`?
so 
`display_context && this.display_context.toUpperCase() === 'READONLY')`

mohammedanas: Unfortunately not,
display_context && this.display_context.toUpperCase() === 'READONLY' would return an undefined.
Its weird how Javascript works., hence the change

PR #359, opened 2019-05-23
Document links validation added and a little refactoring of moving co…

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-3480

Creating a new PR as the previous PR has gone old

https://github.com/hmcts/ccd-case-ui-toolkit/pull/258

We created a uploadDocument event and defined a documentsUpload field and made it mandatory for the uploadDocument event.
Each document uploaded is a complex type of three fields (Type, Document Url and Comment).
But it seems that we can submit the event without selecting a document to upload!
Do you know if this is a known issue with CCD.
 
```
[ ] Yes
[X] No
```


src/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -43,6 +43,11 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
   ngOnInit() {
     this.initDialog();
     let document = this.caseField.value;
+    if (this.caseField.display_context && this.caseField.display_context === Constants.MANDATORY) {
```
rafalkalita: This looks like a logic added. Please add the test for that.

src/shared/commons/constants.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable } from '@angular/core';
+
+@Injectable()
+export class Constants {
```
rafalkalita: I am not sure but maybe we need to add Constants to case-ui-toolkit.module.ts - please run tests and demo locally. If all is good please ignore this comment.

mohammedanas: I have testing this locally in case management and it works all fine

PR #362, opened 2019-05-24
RDM-4622

ganganna-peddapothu:
### Page Show Conditions not working as expected in CaseEventToFields ###

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit/case-edit.component.spec.ts
+++ b/src/shared/components/case-editor/case-edit/case-edit.component.spec.ts
@@ -312,7 +312,7 @@ describe('CaseEditComponent', () => {
 
           expect(wizard.nextPage).toHaveBeenCalled();
           expect(routerStub.navigate).toHaveBeenCalled();
-          expect(component.form.get('data').get(CASE_FIELD_1.id)).toBeNull();
+          expect(component.form.get('data').get(CASE_FIELD_1.id)).not.toBeNull();
           expect(component.form.get('data').get(CASE_FIELD_2.id)).not.toBeNull();
         });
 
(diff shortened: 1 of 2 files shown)


src/shared/services/fields/fields.purger.ts
```diff
@@ -35,10 +36,10 @@ export class FieldsPurger {
     currentPage.wizard_page_fields.forEach(wpf => {
       let case_field = this.findCaseFieldByWizardPageFieldId(currentPage, wpf);
       if (this.hasShowConditionField(case_field, formFields)) {
-          let condition = new ShowCondition(case_field.show_condition);
-          if (this.isHidden(condition, formFields)) {
-            this.resetField(form, case_field);
-          }
+        let condition = new ShowCondition(case_field.show_condition);
+        if (this.isHidden(condition, formFields) && !(this.isReadonly(case_field))) {
```
mario-paniccia: can you move isReadonly method to CaseField class itself please. And add a unit test for it.

src/shared/components/case-editor/case-edit/case-edit.component.spec.ts
```diff
@@ -312,7 +312,7 @@ describe('CaseEditComponent', () => {
 
           expect(wizard.nextPage).toHaveBeenCalled();
           expect(routerStub.navigate).toHaveBeenCalled();
-          expect(component.form.get('data').get(CASE_FIELD_1.id)).toBeNull();
```
mario-paniccia: we should add a separate unit test or 2 or more in this class with proper descriptions, for example:
`it should not clear hidden fields if readonly when leaving the page` or something like that

src/shared/services/fields/fields.purger.ts
ganganna-peddapothu: Added comment on the function signature.

src/shared/components/case-editor/case-edit/case-edit.component.spec.ts
ganganna-peddapothu: Added couple of unit tests for "do not clear READONLY hidden field value" scenario for nextPage and previous page.

veeresha-hmcts:
Verified and ready for merge.

PR #369, opened 2019-05-30
RDM-4264: DAC - Duplicate postcode input IDs

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-4264

### Change description ###
Fixing duplicate IDs for postcode input field, if multiple postcode input fields present

```
[ ] Yes
[x] No
```


src/shared/components/palette/address/write-address-field.html
```diff
@@ -4,20 +4,20 @@
     <h2 class="heading-h2">{{caseField | ccdFieldLabel }}</h2>
 
     <div class="form-group bottom-30" id="postcodeLookup" [ngClass]="{'form-group-error': missingPostcode}" *ngIf="!isExpanded">
```
mario-paniccia: I see other `id` in the html, shouldn't we prefix all of them with id()_?

src/shared/components/palette/address/write-address-field.html
```diff
@@ -4,20 +4,20 @@
...
-      <label for="postcodeInput">
+      <label [for]="id() + '_postcodeInput'">
         <span class="form-label">Enter a UK postcode</span>
       </label>
-        <span class="error-message" *ngIf="missingPostcode">Enter the Postcode</span>
-        <input type="text" id="postcodeInput" name="postcode" class="form-control postcodeinput" [formControl]="postcode">
+      <span class="error-message" *ngIf="missingPostcode">Enter the Postcode</span>
+      <input type="text" [id]="id() + '_postcodeInput'" name="postcode" class="form-control postcodeinput" [formControl]="postcode">
```
mario-paniccia: there's lot of duplication in the creation of the id. Rather that manually concatenating everywhere `id()+`
we should create a method in the component called `createId(string)` and call the method here:
createId(postcodeInput), createId(addressList)

src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -74,8 +76,8 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
   }
 
   focusAddressList() {
-    if (document.getElementById('addressList')) {
-      document.getElementById('addressList').focus();
+    if (document.getElementById(this.idFocus + '_addressList')) {
+      document.getElementById( this.idFocus + '_addressList').focus();
```
mario-paniccia: why we need to introduce `this.idFocus`? 
Can't we just call `this.id()` here?

MKaramath: `this.id()` was failing the unit tests due to some specific browser compatibility problems it seems, so this was the only solution I found,  any suggestions please welcome

src/shared/components/palette/address/write-address-field.html
MKaramath: That's a good point, but this can be taken as a separate tech debt ticket, since this links with several fragile unit test cases, and this may have impact in several other places as well.

src/shared/components/palette/address/write-address-field.html
MKaramath: Done.

src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -74,8 +76,8 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
   }
 
   focusAddressList() {
-    if (document.getElementById('addressList')) {
-      document.getElementById('addressList').focus();
+    if (document.getElementById(this.idFocus + '_addressList')) {
```
mario-paniccia: we must remove document. 
it's not the recommended Angular way and can create other sort of issues. And it's causing unit tests to fails in a flaky way

src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -43,6 +44,7 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
       this.addressList = new FormControl('');
       this.addressFormGroup.addControl('address', this.addressList);
     }
+    this.idFocus = this.createId('addressList');
```
mario-paniccia: rename this to `addressListId` please

src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -128,6 +130,10 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
     return this.isCompoundPipe.transform(field) ? `${this.idPrefix}_` : `${this.idPrefix}`;
   }
 
+  createId(idField: string): string {
```
mario-paniccia: rename `idField` to `elementId` please

MKaramath: Done.

src/shared/components/palette/address/write-address-field.html
MSancaktutar: If that wouldn't take a lot of effort, maybe making this occasional improvement within this ticket can save later context switch cost from the person who will pick a separate ticket in that regard.

PR #372, opened 2019-05-30
Fix show hide condition issue with new collections table and nested collections

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4673

Pre-condition:
Table View (new collection view) is enabled for normal and complex type of elements.
Steps:
    Create a case that has collection field with complex and normal fields. 
    Two collection fields should have Show Condition set 
        CASE1: A collection of a complex type (a set of standard fields and some fields have show conditions )
        CASE2: A collection of a complex type of a complex type (a set of standard fields and some fields have show conditions )
    Submit the case with and without satisfying show conditions

Issue:
CASE1: All hidden fields displayed
CASE2: Show condition field is not displayed
```
[ ] Yes
[X] No
```


veeresha-hmcts:
Verified and ready for merge

PR #374, opened 2019-05-30
Rdm-4535 new contract fix

noronhaa:
Rdm-4535 

### Change description ###
Fix for an issue where pagination on the **Search** result table is not working properly

```
[ ] Yes
[x] No
```


src/shared/components/search-filters/search-filters.component.ts
```diff
@@ -90,7 +107,21 @@ export class SearchFiltersComponent implements OnInit {
     this.selected.page = 1;
     this.selected.metadataFields = this.getMetadataFields();
     this.populateValuesInLocalStorage();
-    this.onApply.emit(this.selected);
+    // Save filters as query parameters for current route
+    let queryParams = {};
```
mario-paniccia: clear duplicated block of code. We should remove this duplication

adnanakgun: Done

PR #375, opened 2019-06-03
RDM-4781-Data not being shown on multi-select list

ganganna-peddapothu:
### Data not being shown on multi-select list using AND and CONTAINS logic ###

```
[ ] Yes
[X ] No
```

### DIFF
--- a/src/shared/services/fields/fields.purger.ts
+++ b/src/shared/services/fields/fields.purger.ts
@@ -13,18 +13,23 @@ export class FieldsPurger {
     private fieldsUtils: FieldsUtils,
   ) {}
 
-  clearHiddenFields(form, wizard, eventTrigger, currentPageId) {
+  clearHiddenFields(form, wizard, eventTrigger, currentPageId, operation) {
     this.clearHiddenFieldForFieldShowCondition(currentPageId, form, wizard, eventTrigger);
-    this.clearHiddenFieldForPageShowCondition(form, wizard);
+    this.clearHiddenFieldForPageShowCondition(currentPageId, form, wizard, operation);
   }
 
-  private clearHiddenFieldForPageShowCondition(form, wizard) {
+  private clearHiddenFieldForPageShowCondition(currentPageId, form, wizard, operation) {
     let formFields = form.getRawValue();
+    let currentPageIndex = this.findIndex(currentPageId, wizard);
+    let isEligibleToResetPage = true;
     wizard.pages.forEach(wp => {
       if (this.hasShowConditionPage(wp, formFields)) {
           let condition = new ShowCondition(wp.show_condition);
-          if (this.isHidden(condition, formFields)) {
+          let wizardPageIndex = this.findIndex(wp.id, wizard);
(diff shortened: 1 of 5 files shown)


ganganna-peddapothu:
> this solution could work. But I think it relies on the fact that we always navigate to the previous page. Which is something we might change in the future if we add the ability to navigate forward to selected pages.
> I think the issue here is that we are passing to the page show condition evaluation only the form fields.
> We should instead pass both the form fields and the event fields. This way the show condition has available also the initial values of the fields. Which now is missing and therefore causing the show condition not to be properly evaluated

ganganna-peddapothu:
Addressed

src/shared/services/fields/fields.purger.ts
```diff
@@ -94,4 +94,8 @@ export class FieldsPurger {
   private isReadonly(case_field: CaseField) {
     return case_field.display_context.toUpperCase() === 'READONLY'
   }
+
+  private findIndex(wizardPageId, wizard) {
```
mario-paniccia: leftover?

src/shared/services/fields/fields.utils.ts
```diff
@@ -127,6 +127,9 @@ export class FieldsUtils {
       return page.parsedShowCondition.match(currentState);
     };
   }
+  public getCurrentEventState(eventTrigger, form): any {
```
mario-paniccia: can you change the methods in this class now to call getCurrentEventState? So we remove duplication

src/shared/services/fields/fields.purger.ts
```diff
@@ -37,23 +37,23 @@ export class FieldsPurger {
...
+        if (this.isHidden(condition, formFields.data) && !(this.isReadonly(case_field))) {
           this.resetField(form, case_field);
         }
       }
     });
   }
 
   private isHidden(condition, formFields) {
-    return !condition.match(formFields.data);
+    return !condition.match(formFields);
```
mario-paniccia: what's the reason of this change?

src/shared/services/fields/fields.purger.ts
```diff
@@ -37,23 +37,23 @@ export class FieldsPurger {
...
   }
 
   private findCaseFieldByWizardPageFieldId(currentPage, wizardPageField) {
     return currentPage.case_fields.find(cf => cf.id === wizardPageField.case_field_id);
   }
 
   private hasShowConditionPage(wizardPage, formFields): boolean {
-    return wizardPage.show_condition && formFields.data[this.getShowConditionKey(wizardPage.show_condition)];
+    return wizardPage.show_condition && formFields[this.getShowConditionKey(wizardPage.show_condition)];
```
mario-paniccia: what's the reason of this change?

ganganna-peddapothu: We are directly getting data fields. We don't have to call again data property.

mario-paniccia: ok got it cheers :)

PR #378, opened 2019-06-05
RDM-4927 | Fixed list not populating values for collection fields

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4927

When trying to use fixed list in collections , the data is not getting populated in the drop down, and also does not display the data in check your answers.

Part of this is due to re instantiating of CaseField object  in the Collections Component.
Possible fix would be to Hyderate the CaseField object with plainToClassFromExist of the class transformer

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/shared/components/palette/base-field/field-read.component.ts
+++ b/src/shared/components/palette/base-field/field-read.component.ts
@@ -1,15 +1,9 @@
-import {
-  Component,
-  ComponentFactoryResolver,
-  Injector,
-  Input, OnInit,
-  ViewChild,
-  ViewContainerRef
-} from '@angular/core';
+import { Component, ComponentFactoryResolver, Injector, Input, OnInit, ViewChild, ViewContainerRef } from '@angular/core';
 import { PaletteService } from '../palette.service';
 import { AbstractFieldReadComponent } from './abstract-field-read.component';
 import { CaseField } from '../../../domain/definition/case-field.model';
 import { FormGroup } from '@angular/forms';
+import { plainToClassFromExist } from 'class-transformer';
 
 @Component({
   selector: 'ccd-field-read',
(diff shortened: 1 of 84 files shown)


src/shared/components/palette/base-field/field-read.component.ts
```diff
@@ -39,7 +33,7 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
     let component = this.resolver.resolveComponentFactory(componentClass).create(injector);
 
     // Provide component @Inputs
-    component.instance['caseField'] = this.caseField;
+    component.instance['caseField'] = plainToClassFromExist(new CaseField(), this.caseField);
```
mario-paniccia: where is this.caseField coming from? wouldn't it be better to apply plainToClassFromExist at the origin rather than here?

mario-paniccia: as you noticed, the field is coming from:
```<table *ngIf="caseField.value && caseField.value.length" class="collection-field-table">
  <tbody>
    <tr *ngFor="let item of caseField.value; let i = index">
      <td>
        <ccd-field-read [caseField]="{
          id: i,
          label: caseField.label + ' ' + (i + 1),
          field_type: caseField.field_type.collection_field_type,
          value: item.value,
          hidden: caseField.hidden
        }" [context]="context"></ccd-field-read>
      </td>
    </tr>
  </tbody>
</table>
```

We should refactor there. Various solutions we can pick the easier. You could create a method in the component that creates the field and call the method in the html.
Or see if it's possible to use plainToClassFromExist or a variant in the html. 
Or use new Field in the html 
etc...

mario-paniccia: It looks like Mohammed did try the above suggestions of mine, but it's not that easy to make it work. So considering we need to release this because of time constraints, we can accept the proposed workaround

PR #381, opened 2019-06-07
RDM-4890 changes to postcode lookup service

mrganeshraja:

https://tools.hmcts.net/jira/browse/RDM-4890

### Change description ###
Change to new postcode lookup service.

```
[x] Yes
[ ] No
```


src/shared/services/addresses/addresses.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    addressesService = injector.get(AddressesService);
+    httpMock = injector.get(HttpTestingController);
+    httpTestingController = TestBed.get(HttpTestingController);
+  });
+
+  it('should be created with all the dependencies', () => {
+    expect(addressesService).toBeTruthy();
+  });
+
+  it('should return atleast more than one addresses from a given postcode', () => {
```
andrewfolga: What is the purpose of this test if we provide the `should return all addresses from a given postcode location`? I would remove it and go with the latter one.

src/shared/services/addresses/addresses.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    const result = addressesService.getAddressesForPostcode(validPostCode);
+    result.subscribe(i => { expect(i.length).toBeGreaterThan(2)});
+  });
+
+  it('should return all addresses from a given postcode location', () => {
+    const result = addressesService.getAddressesForPostcode(validPostCode);
+    result.subscribe(i => { expect(i.length).toEqual(12)});
```
andrewfolga: Sorry to be picky but can we avoid naming variables with single letter? I can see bellow we have `addresses` being used.

src/shared/services/addresses/addresses.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  it('should return addresses with postcode value', () => {
+    const result = addressesService.getAddressesForPostcode(validPostCode);
+    result.subscribe(addresses => { expect(addresses[0].PostCode.length).toBeGreaterThan(0)});
+  });
+
+  it('should return subscriber error when postcode service returns zero', () => {
+    httpService.get.and.returnValue(Observable.of([]));
+    const result = addressesService.getAddressesForPostcode(validPostCode);
+    result.subscribe(addresses => expect(addresses.length).toBe(1),
```
andrewfolga: Why are we expecting addresses to be 1 if there is no values returned from `httpService`?

src/shared/services/addresses/addresses.service.spec.ts
mrganeshraja: I am expecting all the addresses are parsed from the response however, to make it clear I have renamed the tests.

src/shared/services/addresses/addresses.service.spec.ts
mrganeshraja: This is a false assertion.

src/shared/services/addresses/addresses.service.spec.ts
mrganeshraja: Sorry my mistake, I have renamed it now.

src/shared/services/addresses/addresses.service.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  beforeEach(() => {
+
+    appConfig = jasmine.createSpyObj<AbstractAppConfig>('AppConfig', ['getPostcodeLookupUrl',
+      'getPostcodeLookupApiKey']);
+    httpService = jasmine.createSpyObj<HttpService>('HttpService', ['get']);
+    appConfig.getPostcodeLookupUrl.and.returnValue('http://postcodeUrl/postcode=${postcode}&key=${key}');
+    appConfig.getPostcodeLookupApiKey.and.returnValue('apikey');
+    let postCodeResponse = new Response(<ResponseOptions>{body: JSON.stringify(data)});
+    httpService.get.and.returnValue(Observable.of(postCodeResponse));
+    // httpService.get.and.returnValue(Observable.of([]));
```
andrewfolga: Is this needed?

demo/src/app/app.config.ts
```diff
@@ -5,13 +5,15 @@ import { AbstractAppConfig, CaseEditorConfig } from '@hmcts/ccd-case-ui-toolkit'
 @Injectable()
 export class AppConfig extends AbstractAppConfig {
 
+
```
mario-paniccia: can you remove please

src/shared/components/palette/address/address-option.model.ts
```diff
@@ -7,11 +7,12 @@ export class AddressOption {
   constructor (addressModel: AddressModel, description: string) {
     this.value = addressModel;
     this.description = (description == null)
-        ? this.value.AddressLine1
+        ? (this.value.AddressLine1
             +  this.prefixWithCommaIfPresent(this.value.AddressLine2)
```
mario-paniccia: minor: perhaps better extract this block into a private method?

src/shared/services/addresses/addresses.service.ts
```diff
@@ -15,9 +16,40 @@ export class AddressesService {
...
-        map(response => response.json())
-      );
+        map(res=> res.json().results)
+       ).pipe(
+        map(output => output.map(addresses=>
+            AddressesService.mapToAddressModel(addresses["DPA"]))
+      ));
   }
+
+  static mapToAddressModel(address:any) {
```
mario-paniccia: perhaps just a private non static method would do?

src/shared/services/addresses/addresses.service.ts
```diff
@@ -15,9 +16,40 @@ export class AddressesService {
...
+      `${address.BUILDING_NAME} ${address.SUB_BUILDING_NAME} ${address.BUILDING_NUMBER} ${address.THOROUGHFARE_NAME}`);
+    addressModel.AddressLine3 = AddressesService.tidyEmptyValues(
+      `${address.DEPENDENT_LOCALITY} ${address.DOUBLE_DEPENDENT_LOCALITY} ${address.DEPENDENT_THOROUGHFARE_NAME}`);
+    addressModel.PostCode = address.POSTCODE;
+    addressModel.PostTown = address.POST_TOWN;
+    addressModel.Country = 'UK';
+    return addressModel;
+  }
+
+  static tidyEmptyValues(line: String) {
```
mario-paniccia: perhaps better a non static private method?

src/shared/components/palette/address/address-option.model.ts
```diff
@@ -7,11 +7,12 @@ export class AddressOption {
...
               +  this.prefixWithCommaIfPresent(this.value.AddressLine3)
-                + ', ' + this.value.PostTown
+                + ', ' + this.value.PostTown)
                   : description;
+    this.description = this.description.replace(new RegExp("^,", "gi"),'');
```
mario-paniccia: for my knowledge, what's this line for?

src/shared/services/addresses/addresses.service.ts
```diff
@@ -15,9 +16,40 @@ export class AddressesService {
...
+    let addressLine = AddressesService.tidyEmptyValues(
+      `${address.ORGANISATION_NAME} ${address.DEPARTMENT_NAME} ${address.PO_BOX_NUMBER}`)
+    if (addressLine != '') addressModel.AddressLine1 = addressLine;
+    addressModel.AddressLine2 = AddressesService.tidyEmptyValues(
+      `${address.BUILDING_NAME} ${address.SUB_BUILDING_NAME} ${address.BUILDING_NUMBER} ${address.THOROUGHFARE_NAME}`);
+    addressModel.AddressLine3 = AddressesService.tidyEmptyValues(
+      `${address.DEPENDENT_LOCALITY} ${address.DOUBLE_DEPENDENT_LOCALITY} ${address.DEPENDENT_THOROUGHFARE_NAME}`);
+    addressModel.PostCode = address.POSTCODE;
+    addressModel.PostTown = address.POST_TOWN;
+    addressModel.Country = 'UK';
```
mario-paniccia: can you introduce a constant for 'UK' please

src/shared/components/palette/address/address-option.model.ts
mario-paniccia: shall we move this to a method called `removeInitialComma`?

src/shared/services/addresses/addresses.service.ts
```diff
@@ -15,9 +16,40 @@ export class AddressesService {
...
+    return addressModel;
+  }
+
+  static tidyEmptyValues(line: String) {
+    return line
+      .replace(" null", ' ')
+      .replace("null ", ' ')
+      .replace(new RegExp("undefined","gi"), '')
```
mario-paniccia: I'd extract these regex into separate private methods with descriptive names like `removeEmptySpaces`, `removeInitialComma` etc...

src/shared/services/addresses/addresses.service.ts
```diff
@@ -15,9 +16,40 @@ export class AddressesService {
...
-      .get(this.appConfig.getPostcodeLookupUrl().replace('${postcode}', postcode))
+      .get(this.appConfig.getPostcodeLookupUrl()
+        .replace('${postcode}', postcode))
       .pipe(
-        map(response => response.json())
-      );
+        map(res=> res.json().results)
+       ).pipe(
+        map(output => output.map(addresses=>
+            AddressesService.mapToAddressModel(addresses["DPA"]))
```
mario-paniccia: what's this magic keyword `DPA` ? Perhaps better extract it to a constant with a more meaningful name?

demo/src/app/app.config.ts
mrganeshraja: 👍

src/shared/components/palette/address/address-option.model.ts
mrganeshraja: 👍

src/shared/components/palette/address/address-option.model.ts
mrganeshraja: 👍

src/shared/services/addresses/addresses.service.ts
mrganeshraja: 👍

src/shared/services/addresses/addresses.service.ts
mrganeshraja: 👍

src/shared/services/addresses/addresses.service.ts
mrganeshraja: 👍

src/shared/services/addresses/addresses.service.ts
mrganeshraja: 👍

src/shared/components/palette/address/address-option.model.ts
```diff
@@ -1,21 +1,32 @@
...
   constructor (addressModel: AddressModel, description: string) {
-    this.value = addressModel;
-    this.description = (description == null)
-        ? this.value.AddressLine1
-            +  this.prefixWithCommaIfPresent(this.value.AddressLine2)
-              +  this.prefixWithCommaIfPresent(this.value.AddressLine3)
-                + ', ' + this.value.PostTown
-                  : description;
+    if(description == null) {
+      this.value = addressModel;
```
mario-paniccia: we are setting the addressModel only if `description==null`. Before we were not. Is that what we want?

src/shared/services/addresses/addresses.service.ts
```diff
@@ -1,23 +1,69 @@
...
+import { AddressModel } from '../../domain/addresses';
 import { HttpService } from '../http';
+import {map} from "rxjs/operators";
+import 'rxjs/add/observable/of';
 
 @Injectable()
 export class AddressesService {
 
+  private static readonly DPA = 'DPA';
+
```
mario-paniccia: can you remove this empty line please

demo/src/app/app.config.ts
```diff
@@ -11,7 +11,7 @@ export class AppConfig extends AbstractAppConfig {
     'document_management_url': '/documents',
     'login_url': '/login',
     'oauth2_client_id': 'ccd_gateway',
-    'postcode_lookup_url': '/addresses?postcode=${postcode}',
+    'postcode_lookup_url': 'https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode?postcode=${postcode}&key=${key}',
```
mario-paniccia: is this going through the CCD Gateway yes?

src/shared/components/palette/address/address-option.model.ts
```diff
@@ -1,21 +1,34 @@
...
+      this.description = this.getDescription();
+    } else {
+      this.description = description;
```
mario-paniccia: are we forgetting to set the 
`this.value = addressModel;`
in this branch?

src/shared/components/palette/address/write-address-field.component.ts
```diff
@@ -29,11 +30,16 @@ export class WriteAddressFieldComponent extends AbstractFieldWriteComponent impl
 
   addressOptions: AddressOption[];
 
-  missingPostcode = false;
+  alertService: AlertService;
 
-  constructor (addressesService: AddressesService, private isCompoundPipe: IsCompoundPipe) {
+
+  missingPostcode = false
+
```
mario-paniccia: can we avoid multiple blank lines please?

demo/src/app/app.config.ts
mrganeshraja: Yes this is demo app which I had configured to run locally without the gateway. May be a good idea it to go through gateway as well, I will fix it.

src/shared/components/palette/address/address-option.model.ts
mrganeshraja: when the description is passed through constructor it is only used to hold default description that has a total address label on drop down field, We can set it but will never be used.
-----------------------------
 total 12 address found            |   <--------- default option value (show the total count )
-----------------------------
1, Bond street, London W15L
2, Bond Street, London W15L

src/shared/components/palette/address/write-address-field.component.ts
mrganeshraja: 👍  done

src/shared/services/addresses/addresses.service.ts
```diff
@@ -43,14 +42,14 @@ export class AddressesService {
   }
 
   private shiftAddressLinesUp(addressModel: AddressModel) {
-    if (addressModel.AddressLine1 === '') {
-      addressModel.AddressLine1 = addressModel.AddressLine2;
-      addressModel.AddressLine2 = '';
-    }
     if (addressModel.AddressLine2 === '') {
```
mario-paniccia: I've found a little issue int the case line 1 is empty, and line 2 and 3 have a value. This fix will push up line 2 to line 1. But leave line 3 there. It won't be pushed on line 2. not sure if this might be an issue.

veeresha-hmcts:
Verified locally and on Demo

PR #382, opened 2019-06-12
RDM-4996, RDM-4525 As a user I am able to determine the order options in a list

mohammedanas:

https://tools.hmcts.net/jira/browse/RDM-4525

As a user

I want to be able to configure the order of the options in a list

So that I can display certain options above others

Lists in scope:

    fixedlist
    multiselectlist
    fixedRadiolist

 

Add 'DisplayOrder' column to 'FixedList' tab

If DisplayOrder values are given then the options in the lists should appear as per values (1 being first and so on)

If No DisplayOrder is given then order of list is ordered as it is currently

If No DisplayOrder defined then importing is still successful

If I have given DisplayOrder for an list item of 1 and another 10, those items should be displayed in that order and the others displayed as is

Each list by ID can have its own order

Allowing services to override the list order via callback is still allowed and should work as is

Lists can be ordered when used in:

    In a simple field type
    Complex field
    Collection
    Nested Complex Field

```
[ ] Yes
[X] No
```


src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -9,6 +9,7 @@ import { catchError, map, tap } from 'rxjs/operators';
 import { CaseEventData, CaseEventTrigger, CasePrintDocument, CaseView, Draft } from '../../../domain';
 import { WizardPage } from '../domain';
 import { WizardPageFieldToCaseFieldMapper } from './wizard-page-field-to-case-field.mapper';
+import { _ } from 'underscore';
```
andrewfolga: Why are we importing this?

src/shared/services/order/order.service.spec.ts
```diff
@@ -67,6 +67,26 @@ describe('OrderService', () => {
   });
```
andrewfolga: Can we add a test where some of the fixed list field types do and some do not have order with them please? Like the example we discussed today with Aman:
`DD/01/YYYY        regionalCentreEnum    CARDIFF    Cardiff    5
DD/01/YYYY        regionalCentreEnum    MANCHESTER    Manchester    2
DD/01/YYYY        regionalCentreEnum    OXFORD    Oxford`

src/shared/components/case-editor/services/cases.service.ts
mohammedanas: I guess this was already removed in the tsLint changes , may be it did not reflect it

PR #383, opened 2019-06-15
RDM-5053 - Fix bug in conditional show behaviour.

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-5053

1st issue:

- Dynamic List(DivorceCostsOptionDN) rendered based on mid event callback
    Next page depends on the code selected in the dynamic list and the page is never rendered even though the page show condition matches (DivorceCostsOptionDN= "differentAmount" )

2nd issue:

-  Dynamic List(DivorceCostsOptionDN) rendered based on mid event callback

    In the same screen, a textarea field depends on the code selected in the dynamic list and the field is getting rendered when  the page show condition matches (DivorceCostsOptionDN= "differentAmount" ) but upon navigating to next page it will show up only in the second page and in next coming screens, this text area field is lost in the event data.

```
[ ] Yes
[X] No
```


mohammedanas:
As per conference call with @mario-paniccia , it was discussed that the proposed solution in the comments above and other relative solutions that were applied had knockon effects... and it seems the current solution is the most feasible option

PR #385, opened 2019-06-18
RDM-5044 Add Promise then for Angular lifecycle to catchup.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-5044

### Change description ###
Field interpolation is broken on workbasket results

This is ensuring all field values are resolved by label interpolation before the component is fully initialised.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/palette/base-field/field-read.component.ts
+++ b/src/shared/components/palette/base-field/field-read.component.ts
@@ -28,19 +28,19 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
   }
 
   ngOnInit(): void {
-    let componentClass = this.paletteService.getFieldComponentClass(this.caseField, false);
-    let injector = Injector.create([], this.fieldContainer.parentInjector);
-    let component = this.resolver.resolveComponentFactory(componentClass).create(injector);
-
-    // Provide component @Inputs
-    component.instance['caseField'] = plainToClassFromExist(new CaseField(), this.caseField);
-    component.instance['caseFields'] = this.caseFields;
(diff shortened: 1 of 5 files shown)


src/shared/components/palette/base-field/field-read.component.ts
```diff
@@ -28,19 +28,19 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
...
-    // Provide component @Inputs
-    component.instance['caseField'] = plainToClassFromExist(new CaseField(), this.caseField);
-    component.instance['caseFields'] = this.caseFields;
-    component.instance['formGroup'] = this.formGroup;
-    component.instance['caseReference'] = this.caseReference;
-    component.instance['context'] = this.context;
-    component.instance['registerControl'] = this.registerControl || this.defaultControlRegister();
-
-    this.fieldContainer.insert(component.hostView);
+    Promise.resolve(null).then(() => {
```
mario-paniccia: what was the issue? Perhaps we can fix it by using a more appropriate life cycle hook?
https://angular.io/guide/lifecycle-hooks

andrewfolga: The issue is we need the interpolated value inside the ngOnInit to build the field but the value is not there yet. The angular lifecycle kicks in before the label interpolation resolves the placeholder so we need to let it wait a cycle for the already resolved placeholder.

Olawale:
Tested the fixed locally. The field interpolation now display the correct values.

Olawale:
Tested fix on preview. The field interpolation now displays correct value.

PR #387, opened 2019-06-20
Rdm 5107 Collection in Label : First row of a the collection contains values from other rows

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-5107

### Change description ###
Fix field interpolation for placeholders pointing at collection items. Keep the scanning index in current processing context as opposed to zeroing in every new string to resolve processing.

```
[ ] Yes
[x] No
```


src/shared/components/palette/yes-no/write-yes-no-field.html
```diff
@@ -1,4 +1,4 @@
-<div [ngClass]="{'form-group-error': !yesNoControl.valid && yesNoControl.touched}">
+<div [id]="id()" [ngClass]="{'form-group-error': !yesNoControl.valid && yesNoControl.touched}">
```
andrewfolga: Add missing id from yes no write field.

PR #390, opened 2019-06-28
RDM 5031 - redirect to case-list when user do not have access to the case after an event 

smathangi:
https://tools.hmcts.net/jira/browse/RDM-5031

### Change description ###

redirecting to case-list page upon submitting an event when there is a 404 error getting the case.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
+++ b/src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
@@ -77,7 +77,6 @@ export class CaseEventTriggerComponent implements OnInit, OnDestroy {
       .navigate([this.parentUrl], { queryParams: { 'onErrorCaseList': true } })
       .then(() => {
         let caseReference = this.caseReferencePipe.transform(this.caseDetails.case_id.toString());
-        this.alertService.setPreserveAlerts(true);
         if (EventStatusService.isIncomplete(eventStatus)) {
           this.alertService.warning(`Case #${caseReference} has been updated with event: ${this.eventTrigger.name} `
             + `but the callback service cannot be completed`);
--- a/src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
+++ b/src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
@@ -74,9 +74,10 @@ export class CaseEventTriggerComponent implements OnInit, OnDestroy {
   submitted(event: any): void {
     let eventStatus: string = event['status'];
     this.router
-      .navigate([this.parentUrl])
+      .navigate([this.parentUrl], { queryParams: { 'onErrorCaseList': true } })
       .then(() => {
         let caseReference = this.caseReferencePipe.transform(this.caseDetails.case_id.toString());
+        this.alertService.setPreserveAlerts(true);
         if (EventStatusService.isIncomplete(eventStatus)) {
           this.alertService.warning(`Case #${caseReference} has been updated with event: ${this.eventTrigger.name} `
             + `but the callback service cannot be completed`);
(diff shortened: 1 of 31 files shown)


src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
```diff
@@ -74,7 +74,7 @@ export class CaseEventTriggerComponent implements OnInit, OnDestroy {
   submitted(event: any): void {
     let eventStatus: string = event['status'];
     this.router
-      .navigate([this.parentUrl])
+      .navigate([this.parentUrl], { queryParams: { 'onErrorCaseList': true } })
```
smathangi: any better way other than queryParam?

src/shared/components/case-viewer/services/case.resolver.ts
```diff
@@ -87,6 +90,10 @@ export class CaseResolver implements Resolve<CaseView> {
   private checkAuthorizationError(error: any) {
     // TODO Should be logged to remote logging infrastructure
     console.error(error);
+    if (this.onErrorCaseList && error.status === 404) {
+      this.router.navigate(['/list/case'])
```
mario-paniccia: what error are we showing on the workbasket?
does that show as a red banner?

src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
mario-paniccia: mmm I'd avoid introducing a query param. Have you considered checking the previous url? 
https://stackoverflow.com/questions/41038970/how-to-determine-previous-page-url-in-angular

src/shared/components/case-viewer/services/case.resolver.ts
smathangi: as per ticket it's a green confirmation bar on top. Event trigger has these messages already.
https://github.com/hmcts/ccd-case-ui-toolkit/blob/RDM-5031/src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts#L78-L84

src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
smathangi: well on case.resolver.ts i need to know whether it's a redirect to caseList page or standard error page. So, need to pass some flag/property to determine that.

src/shared/components/case-viewer/services/case.resolver.ts
mario-paniccia: 👍

src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
mario-paniccia: what I mean is, on `case.resolver.ts` we check:
 if `error.status === 404` and `coming from submit case url` then redirect to case list

smathangi: done.

src/shared/components/case-viewer/services/case.resolver.ts
```diff
@@ -87,6 +94,10 @@ export class CaseResolver implements Resolve<CaseView> {
   private checkAuthorizationError(error: any) {
     // TODO Should be logged to remote logging infrastructure
     console.error(error);
+    if (CaseResolver.EVENT_REGEX.test(this.previousUrl) && error.status === 404) {
```
mario-paniccia: can we move `CaseResolver.EVENT_REGEX.test(this.previousUrl) && errorStatus === 404` into a private method `cannotAccessUpdatedCase(error.status)`? So it would read:
```
if (!cannotAccessUpdatedCase(error.status)) 
```
a bit more self explanatory I think

PR #392, opened 2019-07-03
RDM-4279 fix for [DAC] Focus Order

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-4279

This PR contains a fix for radio button focus order which occurs in case multiple radio groups are consecutively placed in a wizard page.

```
[ ] Yes
[X] No
```


Olawale:
Tested the fix locally on Chrome and IE11. User can now tab to/focus on content added dynamically.

PR #396, opened 2019-07-09
RDM-4121 - fix undefined error

smathangi:

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/shared/directives/conditional-show/domain/conditional-show.model.ts
+++ b/src/shared/directives/conditional-show/domain/conditional-show.model.ts
@@ -126,8 +126,8 @@ export class ShowCondition {
             try {
               let arrayIndex = Number.parseInt(pathTail[0], 10);
               const [__, ...dropNumberPath] = pathTail;
-              return (fields[head][arrayIndex] !== undefined) ?
-                this.findValueForComplexCondition(fields[head][arrayIndex]['value'], tail[0], tail.slice(1), dropNumberPath.join('_')) : null;
+              return (fields[head][arrayIndex] !== undefined) ? this.findValueForComplexCondition(
+                fields[head][arrayIndex]['value'], tail[0], tail.slice(1), dropNumberPath.join('_')) : null;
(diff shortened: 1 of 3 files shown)


smathangi:
> Sateesh this changes looks good to me. But I was thinking I find this part of the code a bit cryptic.
> Perhaps we can do some little refactoring to make it more understandable? If you can think of any.
> Just pure refactorings to improve code readability. Otherwise let me know and I'll approve as it is
tbh i didn't understand this part of code  - very complex (may be requirement itself complex). Just fixed that exception only 😄. 
I agree with you this definitely needs refactoring - may be another specific task ?

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -126,7 +126,8 @@ export class ShowCondition {
             try {
               let arrayIndex = Number.parseInt(pathTail[0], 10);
               const [__, ...dropNumberPath] = pathTail;
-              return this.findValueForComplexCondition(fields[head][arrayIndex]['value'], tail[0], tail.slice(1), dropNumberPath.join('_'));
+              let newFields = fields[head][arrayIndex] !== undefined ? fields[head][arrayIndex]['value'] : fields[head];
```
rafalkalita: Shouldn't we check why this error is happening and fix the cause? Current solution is just making incorrect data not throwing the error.

smathangi: A click on remove button in turn calls Angular `formArray.removeAt(index)` which is emitting  valueChanges and statusChanges events.  I'm not sure whether we can control these in case of remove only? Talk to you later on this as my UI knowledge is very limited.
https://github.com/hmcts/ccd-case-ui-toolkit/blob/RDM-4121/src/shared/components/palette/collection/write-collection-field.component.ts#L121

In any case above safeValue check helps to deal with bad data rather than blow up.

PR #399, opened 2019-07-11
RDM-4135: Media Viewer integration

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-4135
https://tools.hmcts.net/jira/browse/RDM-5614

### Change description ###
Media viewer Integration with SSR Exclusion

```
[ ] Yes
[x] No
```


src/shared/components/palette/document/media-viewer.component.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  mediaURL: string = '';
+  mediaFilename: string = '';
+  mediaContentType: string = '';
+
+  public constructor(private windowService: WindowService) {
+  }
+
+  ngOnInit() {
+    const localstorageMedia = this.windowService.getLocalStorage(MEDIA_VIEWER);
```
MSancaktutar: localstorageMedia -->  localStorageMedia

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +29,30 @@ export class DocumentManagementService {
...
+    return JSON.stringify(mediaViewer);
+  }
+
+  getContentType(caseField: CaseField): string {
+    let fileExtension = '';
+    if (caseField.value && caseField.value.document_filename) {
+      fileExtension = caseField.value.document_filename
+        .slice(caseField.value.document_filename.lastIndexOf('.') + 1);
+    }
+    return (fileExtension == 'pdf') ? DocumentManagementService.PDF : DocumentManagementService.IMAGE;
```
MSancaktutar: Is it guaranteed that the type is image if it is not pdf? Couldn't it be a third type? This method seems to be problematic.

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +29,30 @@ export class DocumentManagementService {
...
+      }
+    }
+    return JSON.stringify(mediaViewer);
+  }
+
+  getContentType(caseField: CaseField): string {
+    let fileExtension = '';
+    if (caseField.value && caseField.value.document_filename) {
+      fileExtension = caseField.value.document_filename
+        .slice(caseField.value.document_filename.lastIndexOf('.') + 1);
```
MSancaktutar: what if the entire filename is just an extension like "pdf"?

src/shared/services/document-management/document-management.service.spec.ts
```diff
@@ -63,4 +65,51 @@ describe('DocumentManagementService', () => {
         .subscribe(documentMetadata => expect(documentMetadata).toEqual(RESPONSE));
     });
   });
+
+  describe('Media viewer', () => {
```
MSancaktutar: We should have more tests for file names of specific cases. ccd may be allowing only certain types of files but the mediaviewer logic should cover all file types appropriately. The viewer itself, then, may be unable to render some file types, but it should be handled inside the viewer. Our file type extraction logic must not assume certain types only.

tzarsmango: +1 should test for unexpected file types handling as well

tzarsmango: ok, so doc extension addresses other types ;)

src/shared/components/palette/document/media-viewer.component.ts
MKaramath: Done.

src/shared/services/document-management/document-management.service.ts
MKaramath: Refactored by introducing an array with the list of images

src/shared/services/document-management/document-management.service.spec.ts
MKaramath: Added more test cases.

MSancaktutar:
merging to generate prerelease

PR #405, opened 2019-07-25
RDM-5310 Fix for Proceed button does not render correctly

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-5310

This PR fixes two things, 
- Firstly the `contextFields` for the read type complex field renderers (like read-complex-field-table) were reduced to the child field but as dot notation has been introduced the context fields must include all fields so that directives like conditional Hide  and show od legal field substitutor could evaluate correctly
- Fixing this introduced another issue, even though some of the fields were marked as `HIDDEN` they were still being shown in the page. For this `hidden` is honoured for read type complex field renderers.
- Finally since the fields are marked as `READONLY` by the `ccd-data-store-api` when the user has no update write, while adding new collection item the fields weren't rendering correctly. That has been fixed by adding a CollectionCreateCheckerService

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/palette/collection/collection-create-checker.service.ts
+++ b/src/shared/components/palette/collection/collection-create-checker.service.ts
@@ -1,0 +1,41 @@
+import { Injectable } from '@angular/core';
+import { CaseField } from '../../../domain/definition';
+import { Profile } from '../../../domain/profile';
+
+@Injectable()
+export class CollectionCreateCheckerService {
+
+  public setDisplayContextForChildren(caseField: CaseField, caseFields: CaseField[], profile: Profile) {
+    let children = this.getCaseFieldChildren(caseField);
+
+    if (children && children.length > 0) {
+      children.forEach(child => {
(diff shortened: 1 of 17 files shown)


src/shared/components/palette/collection/collection-create-checker.service.ts
```diff
@@ -0,0 +1,41 @@
+import { Injectable } from '@angular/core';
+import { CaseField } from '../../../domain/definition';
+import { Profile } from '../../../domain/profile';
+
+@Injectable()
+export class CollectionCreateCheckerService {
+
+  public setDisplayContextForChildren(caseField: CaseField, caseFields: CaseField[], profile: Profile) {
```
rafalkalita: Shouldn't this be recursive? At the moment it only updates one level below.

tzarsmango: right, added 👍

PR #406, opened 2019-07-25
RDM-5196 change the way search result component handles routing

tzarsmango:
Instead of routing directly the SearchResult component now emits caseId and route must be constructed by the consumer

https://tools.hmcts.net/jira/browse/RDM-5196

```
[ ] Yes
[X] No
```


src/shared/components/search-result/search-result.component.ts
```diff
@@ -48,6 +48,9 @@ export class SearchResultComponent implements OnChanges {
   @Output()
   changePage: EventEmitter<any> = new EventEmitter();
 
+  @Output()
+  clickCase: EventEmitter<any> = new EventEmitter();
```
rafalkalita: It is a new Emitter. Who is listening to that? Or is it something planned for the future?

PR #413, opened 2019-08-08
RDM-5604 Fix {} instantiations for read-collection-XX components

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-5604

CaseField Instantiation with Curly braces causes an issue since curly brace instantiation creates a plain object with no isComplex or isCollection methods. Added a fix for making sure caseField of component is an instance of CaseField by explicit casting...

These are the components using "{}" instantiation
- read-collection-field
- read-complex-field-collection-table

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/palette/base-field/field-read.component.ts
+++ b/src/shared/components/palette/base-field/field-read.component.ts
@@ -28,6 +28,10 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
   }
 
   ngOnInit(): void {
+    // Ensure this.caseField is actually a CaseField instance even if instantiated with {}
+    if (!(this.caseField instanceof CaseField)) {
+      this.caseField = plainToClassFromExist(new CaseField(), this.caseField);
+    }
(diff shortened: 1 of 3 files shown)


src/shared/components/palette/base-field/field-read.component.ts
```diff
@@ -28,6 +28,10 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
   }
 
   ngOnInit(): void {
+    // Ensure this.caseField is actually a CaseField instance even if instantiated with {}
+    if (!(this.caseField instanceof CaseField)) {
+      this.caseField = plainToClassFromExist(new CaseField(), this.caseField);
```
mario-paniccia: 👍 
would the write component have the same issue?

tzarsmango: added to PR description...

Olawale:
I've tested the fix on local docker. The console errors for the unknown methods are are no longer displayed.

PR #421, opened 2019-08-14
RDM-5310 clean rework branch for removing RDM-4073 related changes

tzarsmango:
This PR is for removing RDM-4073 changes from master and previously merged RDM-5310

RDM-4073 will be added back with another PR after extensive tests.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/fixture/case-field-builder.ts
+++ b/src/shared/fixture/case-field-builder.ts
@@ -1,0 +1,84 @@
+import { CaseField, FieldType } from '../domain/definition';
+import { AccessControlList } from '../domain/definition/access-control-list.model';
+
+export class CaseFieldBuilder {
+  private caseField: CaseField = new CaseField();
+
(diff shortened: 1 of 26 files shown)


src/shared/fixture/case-field-builder.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.caseField.show_summary_content_option = option;
+    return this;
+  }
+
+  public withValue(value: any): CaseFieldBuilder {
+    this.caseField.value = value;
+    return this;
+  }
+
+  public withListValue(value: any): CaseFieldBuilder {
```
rafalkalita: Shouldn't it be withListItems and passed an array as a parameter?

tzarsmango: the type of value here is inline with the CaseField.listItems ```  private _list_items: any = [];```

PR #425, opened 2019-08-16
RDM-5594 Collection data lost when parent collection has RD permissions

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-5594

### Change description ###
Changed the table representation of complex type in collections to the old one to fix a data loss issue. New table representation was changing the json structure in Angular controls and not matching with what is stored in the db.

```
[ ] Yes
[ X] No
```


src/shared/components/palette/utils/is-read-only-and-not-collection.pipe.ts
```diff
@@ -19,6 +19,7 @@ export class IsReadOnlyAndNotCollectionPipe implements PipeTransform {
     return this.caseFieldService.isReadOnly(field);
   }
 
+  // CaseField @Expose() doesn't work with the pipe in here, so leaving the manual check
```
andrewfolga: Why this comment here?

rafalkalita: if (field.isCollection()) { has been failing in the console with an error. We use class-transformer to expose isCollection() in CaseField model, but in the above pipe it doesn't work. The comment is there to make a person think before changing it.

src/shared/components/palette/collection/write-collection-field.component.spec.ts
```diff
@@ -525,11 +525,11 @@ describe('WriteCollectionFieldComponent CRUD impact - Update False', () => {
     fixture.detectChanges();
   }));
 
-  it('should render a row with a read field for each items when user does not have update right', () => {
```
andrewfolga: Any chance we keep the new test in the same format so we assert on the things we query from the resulting template rather than on the component itself?

PR #434, opened 2019-09-06
Rdm 4135 mv integration with ssr exclusion

MSancaktutar:
https://tools.hmcts.net/jira/browse/RDM-4135
https://tools.hmcts.net/jira/browse/RDM-5614

### Change description ###
MV Integration with SSR Exclusion

```
[ ] Yes
[x] No
```


src/shared/components/palette/document/read-document-field.component.spec.ts
```diff
@@ -72,7 +89,19 @@ describe('ReadDocumentFieldComponent', () => {
...
+
+    it('should call Media Viewer when the document link is clicked', () => {
+      component.caseField.value = VALUE;
+      fixture.detectChanges();
+      spyOn(component, 'showMediaViewer');
+      let linkElement = de.query(By.css('a'));
+      expect(linkElement).toBeTruthy();
+      linkElement.triggerEventHandler('click', null);
+      fixture.detectChanges();
+      component.showMediaViewer();
```
andrewfolga: Should we not remove this line and assert based on the click on a link only?

tsconfig-aot.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "module": "es2015",
+    "module": "commonjs",
```
andrewfolga: Why are we changing this?

MSancaktutar: Lazy loading is mandating this.

src/shared/components/palette/document/read-document-field.component.ts
```diff
@@ -1,8 +1,30 @@
...
+export class ReadDocumentFieldComponent extends AbstractFieldReadComponent {
+
+  constructor(private windowService: WindowService,
+              private documentManagement: DocumentManagementService,
+              private router: Router) {
+    super();
+  }
+
+  showMediaViewer(): void {
+    let routerMediaViewer = this.router.createUrlTree(['/media-viewer']);
```
mario-paniccia: shall we rename this `mediaViewerUrl`?

mario-paniccia: or we could introduce a private method `getMediaViewerUrl(this.router)`

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +31,40 @@ export class DocumentManagementService {
...
+        document_filename: caseField.value.document_filename,
+        content_type: this.getContentType(caseField),
+      }
+    }
+    return JSON.stringify(mediaViewer);
+  }
+
+  getContentType(caseField: CaseField): string {
+    let fileExtension = '';
+    if (caseField.value && caseField.value.document_filename) {
```
mario-paniccia: `caseField.value` check redundant, it has already been checked in `createMediaViewer` method

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +31,40 @@ export class DocumentManagementService {
         map(response => response.json())
       );
   }
+
+  createMediaViewer(caseField: CaseField): string {
+    let mediaViewer = {};
+    if (caseField.value) {
+      mediaViewer = {
+        document_binary_url: this.transformDocumentUrl(caseField.value.document_binary_url),
```
mario-paniccia: `caseField.value` is duplicated a lot in this new code. Perhaps it'll be leaner to pass `caseField.value` directly to this `createMediaViewer` method rather than `caseField`. And invoke `createMediaViewer` only if `caseField.value` is not null so you don't have to check that in this service class multiple times

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +31,40 @@ export class DocumentManagementService {
...
+    if (caseField.value && caseField.value.document_filename) {
+      fileExtension = caseField.value.document_filename
+        .slice(caseField.value.document_filename.lastIndexOf('.') + 1);
+    }
+    if (this.isImage(fileExtension)) {
+      return DocumentManagementService.IMAGE;
+    } else if (fileExtension === 'pdf') {
+      return DocumentManagementService.PDF;
+    } else {
+      return null;
```
mario-paniccia: shall we consider adding some sort of info/warn logging when the content type is unknown?

src/shared/components/palette/document/read-document-field.component.ts
MKaramath: `.toString()` will fail if the router returns null, so ignoring it for now

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +31,40 @@ export class DocumentManagementService {
         map(response => response.json())
       );
   }
+
+  createMediaViewer(caseField: CaseField): string {
```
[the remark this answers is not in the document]
MKaramath: Renamed to `getMediaViewerInfo`.

src/shared/services/document-management/document-management.service.ts
MKaramath: Removed the check.

src/shared/services/document-management/document-management.service.ts
MKaramath: Refactored by passing `caseField.value` as parameter

src/shared/services/document-management/document-management.service.ts
MKaramath: Done.

src/shared/components/palette/document/read-document-field.component.spec.ts
MKaramath: Removed.

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +30,38 @@ export class DocumentManagementService {
         map(response => response.json())
       );
   }
+
+  getMediaViewerInfo(caseFieldValue: any): string {
```
mario-paniccia: can we rename `caseFieldValue` to `documentFieldValue`?

src/shared/services/document-management/document-management.service.ts
```diff
@@ -26,4 +30,38 @@ export class DocumentManagementService {
...
+      console.warn(`Unknown content type with the file extension: ${fileExtension}`);
+      return null;
+    }
+  }
+
+  isImage(imageType: string) {
+    return this.imagesList.find(e => e === imageType.toUpperCase()) !== undefined;
+  }
+
+  transformDocumentUrl(value: string): string {
```
mario-paniccia: can we rename `value` to `document_binary_url`?

src/shared/components/palette/document/read-document-field.component.ts
mario-paniccia: then we should add a null check. And you are doing that in the called method anyway:
```if (urlTree) {
      window.open(urlTree.toString(), '_blank');
    }```
so better have it here then.

File: src/shared/components/palette/document/read-document-field.component.ts [hunk 10]
```diff
@@ -1,8 +1,29 @@
...
+    this.windowService.removeLocalStorage(MEDIA_VIEWER);
+    if (this.caseField && this.caseField.value) {
+      this.windowService.setLocalStorage(MEDIA_VIEWER, this.documentManagement.getMediaViewerInfo(this.caseField.value));
```
mario-paniccia: can we find a better name for the constant `MEDIA_VIEWER`? Perhaps `MEDIA_VIEWER_INFO` is a bit more clear?

src/shared/services/document-management/document-management.service.ts
MKaramath: Done.

src/shared/components/palette/document/read-document-field.component.ts
MKaramath: Done.

src/shared/components/palette/document/read-document-field.component.ts
MKaramath: Done.

PR #436, opened 2019-09-13
Rdm-5696 - Fix dynamic list values

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-5696

Fix the dynamic list data getting submitted in json object instead of string format.

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -127,8 +127,11 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       let caseEventData: CaseEventData = this.formValueService.sanitise(pageFormFields) as CaseEventData;
       caseEventData.event_token = this.eventTrigger.event_token;
       caseEventData.ignore_warning = this.ignoreWarning;
-      caseEventData.event_data = this.editForm.value.data;
-
+      let event_data = this.editForm.value.data
+      Object.keys(pageFormFields.data).forEach((key) => {
+        event_data[key] = pageFormFields.data[key];
+      });
+      caseEventData.event_data = event_data;
       if (this.caseEdit.caseDetails) {
         caseEventData.case_reference = this.caseEdit.caseDetails.case_id;
       }
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -127,10 +127,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       let caseEventData: CaseEventData = this.formValueService.sanitise(pageFormFields) as CaseEventData;
       caseEventData.event_token = this.eventTrigger.event_token;
       caseEventData.ignore_warning = this.ignoreWarning;
-      let event_data = this.editForm.value.data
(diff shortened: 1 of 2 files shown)


src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -127,8 +127,11 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       let caseEventData: CaseEventData = this.formValueService.sanitise(pageFormFields) as CaseEventData;
       caseEventData.event_token = this.eventTrigger.event_token;
       caseEventData.ignore_warning = this.ignoreWarning;
-      caseEventData.event_data = this.editForm.value.data;
-
+      let event_data = this.editForm.value.data
+      Object.keys(pageFormFields.data).forEach((key) => {
+        event_data[key] = pageFormFields.data[key];
```
mario-paniccia: @mohammedanas one thing. I'm not sure but can you please verify the following: 
You are reading the fields from `pageFormFields`. But `pageFormFields` only has got the current page fields. We need to make sure 
also the dynamic fields in the previous pages are passed correctly to the MidEvent callback. Can you verify that happens please?

mohammedanas: Already thought about it Mario, So as per code its not going to update the previous fields so they remain same.. we deal on per page basis, thats fine as well

mario-paniccia: would it be possible to add a quick unit test to confirm that and avoid regressions? if this doesn't take too much time

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -127,8 +127,11 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       let caseEventData: CaseEventData = this.formValueService.sanitise(pageFormFields) as CaseEventData;
       caseEventData.event_token = this.eventTrigger.event_token;
       caseEventData.ignore_warning = this.ignoreWarning;
-      caseEventData.event_data = this.editForm.value.data;
-
+      let event_data = this.editForm.value.data
```
mario-paniccia: can you move this code into a private method with a good descriptive name please? it's not clear why we are including this code

mohammedanas: Done!

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
mohammedanas: I think, It will be more of setting up scenarios of multiple pages and interactions between them, not sure how we can acheive them, in a unit test. I am open to options, but at this stage and urgency, manual tests are fine.

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -148,6 +148,14 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
     }
   }
 
+  private convertDynamicListFieldsForMidEventCallback(pageFormFields) {
```
MSancaktutar: Could this please be strongly-typed?

hemantt:
Closing the stale PR

PR #439, opened 2019-09-17
RDM-5296 - Table Collection View - Case Link is not displayed as a link

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-5296

### Change description ###
CaseLink was displayed as a text. This change will make it a link.

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/shared/components/palette/complex/read-complex-field-collection-table.html
+++ b/src/shared/components/palette/complex/read-complex-field-collection-table.html
@@ -75,9 +75,7 @@
                                 <span class="text-16"><ccd-field-read [caseField]="field" [context]="context"></ccd-field-read></span>
                               </td>
                               <td *ngIf="name.value.hasOwnProperty('CaseReference')">
-                                <a href="/v2/case/{{name.value.CaseReference}}" target="_blank" class="caselink">
-                                  <span class="text-16">{{name.value.CaseReference ? (transformReference(name.value.CaseReference)) : ''}}</span>
-                                </a>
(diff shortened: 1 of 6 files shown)


src/shared/components/palette/complex/read-complex-field-collection-table.html
```diff
@@ -71,8 +71,13 @@
                             <tr class="complex-panel-nested-field" ccdConditionalShow ccdLabelSubstitutor [caseField]="field"
                                 [contextFields]="vLabel.value.caseField  | ccdFieldsFilter">
                               <th><span class="text-16">{{field.label}}</span></th>
-                              <td>
-                                <span class="text-16"><ccd-field-read [caseField]="field" [context]="context"></ccd-field-read></span>
+                              <td *ngIf="!name.value.hasOwnProperty('CaseReference')">
+                                <span class="text-16"><ccd-field-read [caseField]="addValue(field, name.value)" [context]="context"></ccd-field-read></span>
+                              </td>
+                              <td *ngIf="name.value.hasOwnProperty('CaseReference')">
+                                <a href="/v2/case/{{name.value.CaseReference}}" target="_blank" class="caselink">
```
andrewfolga: Shouldn't we use `case/:jid/:ctid/:cid` rather than `v2/case/:cid`?
The latter will be removed after v2 stories cleanup and the former will remain which is used now when you click through to get to a case from case list.

andrewfolga:
As discussed try reusing `read-case-link-field` component.

PR #446, opened 2019-10-09
Rdm 5696 changes need to cherry picked into this branch

mohammedanas:
Rdm 5696 changes need to cherry picked into this branch

https://tools.hmcts.net/jira/browse/RDM-5696

Dynamic list contents received by call back services are in string format and not in the standard DynamicList format.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -127,8 +127,8 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       let caseEventData: CaseEventData = this.formValueService.sanitise(pageFormFields) as CaseEventData;
       caseEventData.event_token = this.eventTrigger.event_token;
       caseEventData.ignore_warning = this.ignoreWarning;
+      this.updateMainFormWithChangesFromPageForm(this.editForm.value.data, pageFormFields);
       caseEventData.event_data = this.editForm.value.data;
-
       if (this.caseEdit.caseDetails) {
         caseEventData.case_reference = this.caseEdit.caseDetails.case_id;
       }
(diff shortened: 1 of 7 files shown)


src/shared/services/form/field-type-sanitiser.spec.ts
```diff
@@ -20,6 +21,23 @@ describe('FieldTypeSanitiser', () => {
     '            ]\n' +
     '          }');
 
+  const COMPLEX_VALUE_DYNAMIC_LIST = JSON.parse('{\n' +
```
andrewfolga: This is never used. Do we need it?

src/shared/services/form/field-type-sanitiser.ts
```diff
@@ -12,11 +12,22 @@ export class FieldTypeSanitiser {
    * @param editForm
    */
    sanitiseLists(caseFields: CaseField[], editForm: any) {
-
     this.getDynamicListsFromCaseFields(caseFields).forEach(dynamicField => {
-      this.getListOfKeysFromEditForm(editForm).forEach((key) => {
-        this.createValueCodePairAlongWithListIfKeyExistsInForm(dynamicField, key, editForm);
```
andrewfolga: Did we remove the method as it is not used anywhere else and is not being used here now?

src/shared/services/form/field-type-sanitiser.ts
```diff
@@ -12,11 +12,22 @@ export class FieldTypeSanitiser {
...
+        this.getDeepKeys(editForm).filter(key => key.endsWith(dynamicField.id)).forEach((key) => {
+          let formValue = this.getNestedValue(editForm, key)
+          let value = {
+            value: this.getMatchingCodeFromListOfItems(dynamicField, formValue, key),
+            list_items: dynamicField.list_items
+          };
+          this.setValue(key, value, editForm);
+        });
+
+      } else if (dynamicField.field_type.type === 'Collection') {
```
andrewfolga: Didn't we say for SSCS it should be simple dynamic list only (top level)?

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -148,6 +148,12 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
     }
   }
 
+  private updateMainFormWithChangesFromPageForm(editFormFields, pageFormFields) {
```
rafalkalita: As per conversation with Andrew this will only work for the top level (not in complex) field. This is what we do support at the moment and needs to be communicated with the ui-toolkit users.

PR #449, opened 2019-10-13
Rdm 4955

MSancaktutar:
https://tools.hmcts.net/jira/browse/RDM-4955
https://tools.hmcts.net/jira/browse/RDM-6127

### Change description ###
Correction for the issue with TIF files as realised during functional test automations

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/services/document-management/document-management.service.ts
+++ b/src/shared/services/document-management/document-management.service.ts
@@ -13,7 +13,7 @@ export class DocumentManagementService {
   private static readonly PDF = 'pdf';
   private static readonly IMAGE = 'image';
 
-  imagesList: string[] = ['JPEG', 'GIF', 'PNG', 'TIF', 'JPG'];
+  imagesList: string[] = [ 'GIF', 'JPG', 'JPEG', 'PNG'];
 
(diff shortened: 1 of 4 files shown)


src/shared/services/document-management/document-management.service.ts
```diff
@@ -13,7 +13,7 @@ export class DocumentManagementService {
   private static readonly PDF = 'pdf';
   private static readonly IMAGE = 'image';
 
-  imagesList: string[] = ['JPEG', 'GIF', 'PNG', 'TIF', 'JPG'];
+  imagesList: string[] = [ 'GIF', 'JPG', 'JPEG', 'PNG'];
```
mario-paniccia: @MSancaktutar wouldn't it better to have this list as a config property rather than hardcoded?

MSancaktutar: Yes, indeed it would be. But the original code had been merged with hard-coded list of accepted image extensions, and I just wouldn't want to go beyond the purpose of RDM-4955 in this one. I opened RDM-6099 for this, but Media Viewer's parameters may change in later releases, and it may turn this ticket obsolete.

PR #452, opened 2019-10-15
Rdm 1280 merged with rdm 5398

Thor-tech-of-metal:
RDM-1280: Wizard page should only allow Continue when all documents have been uploaded

    [RDM-1280](https://tools.hmcts.net/jira/browse/RDM-1280).

Notes: This tickets depends on  [RDM-5709](https://tools.hmcts.net/jira/browse/RDM-5709).
It can not be merged until RDM-5709 is merged. 

The same regressions test of RDM-5709 should be done.

### DIFF
--- a/src/shared/components/palette/document/write-document-field.component.ts
+++ b/src/shared/components/palette/document/write-document-field.component.ts
@@ -17,6 +17,8 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
   static readonly DOCUMENT_FILENAME = 'document_filename';
   static readonly UPLOAD_ERROR_FILE_REQUIRED = 'File required';
   static readonly UPLOAD_ERROR_NOT_AVAILABLE = 'Document upload facility is not available at the moment';
+  static readonly UPLOAD_WAITING_FILE_STATUS = 'Uploading...';
+
   private uploadedDocument: FormGroup;
   private selectedFile: File;
   private dialogConfig: MatDialogConfig;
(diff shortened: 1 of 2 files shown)


src/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -210,4 +212,10 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
 
     return error.error;
   }
+
+  private displayFileWaitingStatus () {
+
```
mario-paniccia: can u remove this empty line please

src/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -210,4 +212,10 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
...
+    this.valid = false;
+    this.uploadError = WriteDocumentFieldComponent.UPLOAD_WAITING_FILE_STATUS;
```
mario-paniccia: are we using the `uploadError` to display generic info messages as well? It's not his purpose. Perhaps we could improve this by either renaming `uploadError` to something more generic, or introducing a new property for displaying info messages so we clearly separate the 2

src/shared/services/document-management/document-management.service.ts
```diff
@@ -22,8 +23,11 @@ export class DocumentManagementService {
     headers.append(DocumentManagementService.HEADER_CONTENT_TYPE, null);
     return this.http
       .post(url, formData, { headers })
+      .pipe(delay( 3000 ))
```
mario-paniccia: move 3000 to a constant and add an explanatory comment on why this has been added please

PR #458, opened 2019-10-25
Rdm 4837 Field interpolation syntax is displayed if the source field is NULL

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4837

### Change description ###
Field interpolation syntax is displayed if the source field is NULL

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/directives/substitutor/services/placeholder.service.ts
+++ b/src/shared/directives/substitutor/services/placeholder.service.ts
@@ -31,7 +31,7 @@ ___
         private fieldIdToSubstitute: string;
         private startSubstitutionIndex: number;
         private isCollecting: boolean;
-        private originalStringToResolve: string;
+        private readonly originalStringToResolve: string;
         constructor(values: {
             stringToResolve: string,
             pageFormFields: any}) {
--- a/src/shared/directives/substitutor/services/placeholder.service.ts
+++ b/src/shared/directives/substitutor/services/placeholder.service.ts
@@ -36,7 +36,7 @@ ___
 
     resolvePlaceholders(pageFormFields, stringToResolve): string {
         let pt = new PlaceholderService.ProgressTracker({pageFormFields: pageFormFields, stringToResolve: stringToResolve});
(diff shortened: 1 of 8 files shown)


src/shared/directives/substitutor/services/placeholder.service.ts
```diff
@@ -18,58 +19,88 @@ ___
       ) { }
 
     resolvePlaceholders(pageFormFields, stringToResolve): string {
```
MSancaktutar: Function way too long and complicated to read. Could you please break into smaller ones?

src/shared/directives/substitutor/services/placeholder.service.ts
```diff
@@ -18,58 +19,91 @@ ___
...
-                                                newNumberOfCollectionItemsAsPlaceholder,
-                                                numberCollectionItemsAsPlaceholder)
-                                    }
-                                    stringToResolve = this.substitute(
-                                        pageFormFields, stringToResolve, startSubstitutionIndex, fieldIdToSubstitute, colItemIndex);
-
-                                    scanIndex = this.resetScanIndexAfterSubstitution(
-                                        startSubstitutionIndex, pageFormFields, fieldIdToSubstitute, colItemIndex);
+
+        while (this.hasUnresolvedPlaceholder(stringToResolve)) {
```
MSancaktutar: This code should further be refactored to have no more than 2 or 3 levels of conditional control in a function.
Appreciate the original code was already problematic in this aspect, and if refactoring now is too much beyond the purpose of the particular change, let's then define a new refactoring ticket just so that we don't miss this tech debt off our future agenda.

andrewfolga: Ticket created: https://tools.hmcts.net/jira/browse/RDM-6248

andrewfolga: Closed the above as was done as part of this ticket now.

PR #460, opened 2019-10-28
Rdm 5303 based rc 15.2

Thor-tech-of-metal:
RDM-5303: Document file name disappears when you come back to the page from a previous page.
    
      [RDM-5303](https://tools.hmcts.net/jira/browse/RDM-5303).

### DIFF
--- a/src/shared/commons/collections-helpers.ts
+++ b/src/shared/commons/collections-helpers.ts
@@ -1,0 +1,14 @@
+export class CollectionsHelpers {
+  constructor() {
+  }
+
+  filterElement(el) {
+    return el != null;
+  }
+
+  flatAnArray(array) {
+    return array.reduce((acc, item) => {
+      return acc.concat(item);
+    }, []);
+  }
+}
--- a/src/shared/commons/collections-helpers.ts
+++ b/src/shared/commons/collections-helpers.ts
@@ -1,10 +1,5 @@
 export class CollectionsHelpers {
-  constructor() {
-  }
-
-  filterElement(el) {
-    return el != null;
-  }
+  constructor() { }
 
   flatAnArray(array) {
     return array.reduce((acc, item) => {
--- a/src/shared/commons/form-group-helper.ts
+++ b/src/shared/commons/form-group-helper.ts
@@ -3,14 +3,15 @@ import { AbstractControl, FormArray, FormGroup } from '@angular/forms';
 export class FormGroupHelper {
   constructor() {}
   /**
-   * This function is going to find a component's value from the top level FormGroup.
-   * @param componentName this is the component name that is going to be used to find the component's value.
-   * @param group this is the top level FormGroup.
-   * @return not-Defined if the component's value could not be found. componentValue:any
-   * if there is a value for the component name passed as parameter.
+   * This function is going to find a component's value by a component id. It will search the component's value
+   * in the FormGroup container.
+   * @param componentId this is the component id that is going to be used to find the component's value.
+   * @param group this is the top level FormGroup container.
+   * @return not-Defined if the component'value could not be found. componentValue:any
+   * if there is a component's value is found.
    */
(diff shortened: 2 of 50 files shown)


src/shared/commons/form-group-helper.ts
```diff
@@ -0,0 +1,69 @@
...
+            return abstractControl.value;
+          }
+        }
+      }
+      return control;
+    }).filter(this.filterElement);
+    const flattenedArray = this.flatAnArray(formControls);
+    return this.flatAnArray(Array.from(new Set(flattenedArray).values()));
+  }
+
```
Thor-tech-of-metal: I am going to create a separated class for all collections helpers methods.

src/shared/commons/collections-helpers.ts
```diff
@@ -0,0 +1,14 @@
+export class CollectionsHelpers {
+  constructor() {
+  }
+
+  filterElement(el) {
```
mario-paniccia: a bit misleading method it's not doing a filtering it's just doing a null check. Also, not operating on a collection. Perhaps better removing it

src/shared/commons/collections-helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+    return el != null;
+  }
+
+  flatAnArray(array) {
```
mario-paniccia: I assume we have already checked whether javasctipt provide a fatten method already right?

Thor-tech-of-metal: Array.prototype.flat is not supported by in our Karma test settings. We need to increase ECMAScript.  It works in the browser but it fails im the yarn test .

src/shared/commons/collections-helpers.ts
Thor-tech-of-metal: i will  remove it.

src/shared/commons/collections-helpers.ts
mario-paniccia: fair enough 👍

src/shared/commons/form-group-helper.ts
```diff
@@ -0,0 +1,63 @@
+import { AbstractControl, FormArray, FormGroup } from '@angular/forms';
+import { CollectionsHelpers } from './collections-helpers';
+
+export class FormGroupHelper {
```
mario-paniccia: mark this as `@Injectable()` so that you can inject it where needed rather than having every time to create new instances. See for example `FieldsUtils`

src/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -190,20 +192,28 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
     this.uploadedDocument.get(WriteDocumentFieldComponent.DOCUMENT_URL).setValue(url);
     this.uploadedDocument.get(WriteDocumentFieldComponent.DOCUMENT_BINARY_URL).setValue(binaryUrl);
     this.uploadedDocument.get(WriteDocumentFieldComponent.DOCUMENT_FILENAME).setValue(filename);
+    const value = {
```
mario-paniccia: let's introduce a private method that gets as parameters `url`, `binaryUrl` and `document_filename` and creates the object

src/shared/commons/form-group-helper.ts
Thor-tech-of-metal: OK

src/shared/components/palette/document/write-document-field.component.ts
Thor-tech-of-metal: OK

src/shared/commons/form-group-helper.ts
Thor-tech-of-metal: if I create a Bean for that I will have change the @Module test coverage  for this project as well as CaseManagement hence another pr and new release for CaseManagement .

PR #461, opened 2019-11-07
RDM-6350 Introduce new ErrorNotifierService and subscribe in CaseViewerComponent

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-6350

### Change description ###
Introduce new ErrorNotifierService that will notify CaseViewerComponent about any error that occurred during navigation done by the consumer of the CaseViewerComponent

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-viewer/case-viewer.component.ts
+++ b/src/shared/components/case-viewer/case-viewer.component.ts
@@ -82,7 +82,7 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
       this.error = errorEvent;
     });
     this.errorSubscription = this.errorNotifierService.error.subscribe(error => {
-      if (error.status && error.status !== 401 && error.status !== 403) {
+      if (error && error.status !== 401 && error.status !== 403) {
         this.error = error;
         this.callbackErrorsSubject.next(this.error);
       }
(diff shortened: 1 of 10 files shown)


src/shared/components/case-viewer/case-viewer.component.ts
```diff
@@ -78,6 +81,12 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
     this.callbackErrorsSubject.subscribe(errorEvent => {
       this.error = errorEvent;
     });
+    this.errorSubscription = this.errorNotifierService.error.subscribe(error => {
+      if (error.status && error.status !== 401 && error.status !== 403) {
+        this.error = error;
```
mario-paniccia: can we have a situation where `error.status` is empty or null and so even though there is an `error` the if it's not executed hence the error not displayed?

src/shared/components/case-viewer/services/case.resolver.ts
```diff
@@ -4,7 +4,7 @@ import { Observable } from 'rxjs';
 import { catchError, map } from 'rxjs/operators';
 import { CaseView, Draft } from '../../../domain';
 import { CaseNotifier, CasesService } from '../../case-editor';
-import { AlertService, DraftService, NavigationOrigin } from '../../../services';
```
andrewfolga: Removed as unused.

PR #465, opened 2019-11-08
Rdm 5735

Thor-tech-of-metal:
RDM-5735: Fixed list display order should be ASC, not DESC

      [RDM-5735](https://tools.hmcts.net/jira/browse/RDM-5735).

### DIFF
--- a/src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
+++ b/src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
@@ -1,6 +1,7 @@
 import { Component, OnInit } from '@angular/core';
 import { FormControl } from '@angular/forms';
 import { AbstractFieldWriteComponent } from '../base-field/abstract-field-write.component';
+import { OrderService } from '../../../services/order';
 
 @Component({
   selector: 'ccd-write-fixed-list-field',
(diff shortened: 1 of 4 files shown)


src/shared/services/order/order.service.ts
```diff
@@ -27,6 +27,23 @@ export class OrderService {
     return a.order - b.order;
   };
 
+  public static sortListItems(caseField: CaseField) {
```
andrewfolga: Can we add tests for cases:
1) we do not have order values specified - final order should reverse the original order of elements
2) order is provided but some elements are missing the order - final order should reverse the original order of elements
3) order is provided for all elements - final order should follow the defined order

andrewfolga: As discussed I understand that 2. is not to be covered by this PR as there will be future validation logic added in def store that will stop users from importing this partial state.
1. and 3. are covered as part of write-fixed-list-field.component.spec.ts.

These tests should really be moved to order.service.ts at some point but a minor comment really.

src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
```diff
@@ -13,6 +14,7 @@ export class WriteFixedListFieldComponent extends AbstractFieldWriteComponent im
   ngOnInit() {
     let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
     this.fixedListControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null));
+    OrderService.sortListItems(this.caseField);
```
mario-paniccia: we should use dependency injection. OrderService is injectable.

Thor-tech-of-metal:
No longer needed

PR #472, opened 2019-11-18
Rdm 4324 DisplayOrder required for ComplexType elements

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4324

### Change description ###
DisplayOrder required for ComplexType elements

```
[ ] Yes
[x] No
```


src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
```diff
@@ -112,6 +112,8 @@ export class ReadComplexFieldCollectionTableComponent extends AbstractFieldReadC
     return result;
   }
 
+  keepOriginalOrder = (a, b) => a.key;
```
rafalkalita: Is it needed? I have removed it and all tests are green.

src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -240,8 +240,8 @@ export class CasesService {
 
     eventTrigger.wizard_pages.forEach((wizardPage: WizardPage) => {
       wizardPage.parsedShowCondition = new ShowCondition(wizardPage.show_condition);
-      wizardPage.case_fields = this.orderService
-        .deepSort(this.wizardPageFieldToCaseFieldMapper.mapAll(wizardPage.wizard_page_fields, eventTrigger.case_fields));
+      wizardPage.case_fields = this.orderService.sort(
```
rafalkalita: do we still need to use orderService?

andrewfolga: Yes the order service still orders the top level fields but the BE is dealing with ordering of compound fields now.

src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
andrewfolga: Yes this is needed to keep original order of the entries from the ordered returned by BE. Otherwise Angular and browser will reorder alphabetically.

andrewfolga: Tests added now.

PR #492, opened 2020-01-06
RDM-7120: Regression: Unable to create Draft SDO in DEMO

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-7120

### Change description ###

Change select element binding to [(ngModel)] and use [ngValue] instead of formControl and "value" attributes

```
[ ] Yes
[X ] No
```

### DIFF
--- a/src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
+++ b/src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
@@ -7,10 +7,23 @@ import { AbstractFieldWriteComponent } from '../base-field/abstract-field-write.
   templateUrl: './write-fixed-list-field.html'
 })
 export class WriteFixedListFieldComponent extends AbstractFieldWriteComponent implements OnInit {
+
+  fixedListFormControl: FormControl;
+
   ngOnInit() {
     let isNull = this.caseField.value === undefined || this.caseField.value === '';
     if (isNull) {
       this.caseField.value = null;
+    }
+    if (this.registerControl && !this.caseField.metadata) {
+      this.fixedListFormControl = new FormControl(this.caseField.value);
+      this.registerControl(this.fixedListFormControl);
+    }
+  }
+
+  onValueChange() {
+    if (this.fixedListFormControl) {
+      this.fixedListFormControl.setValue(this.caseField.value);
     }
   }
 }
--- a/src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
+++ b/src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
@@ -15,15 +15,6 @@ export class WriteFixedListFieldComponent extends AbstractFieldWriteComponent im
     if (isNull) {
       this.caseField.value = null;
     }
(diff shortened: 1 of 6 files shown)


src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
```diff
@@ -7,18 +7,10 @@ import { AbstractFieldWriteComponent } from '../base-field/abstract-field-write.
...
-    if (this.hasSanitisedDynamicListData()) {
-      this.fixedListControl.setValue(this.caseField.value);
+    let isNull = this.caseField.value === undefined || this.caseField.value === '';
+    if (isNull) {
+      this.caseField.value = null;
     }
   }
-
-  private hasSanitisedDynamicListData() {
-    return this.fixedListControl.value && this.fixedListControl.value.value;
```
mario-paniccia: this line seems redundant to me so it should be ok to remove. But just for piece of mind can we verify dynamic lists are still working?

src/shared/components/palette/fixed-list/write-fixed-list-field.html
```diff
@@ -1,16 +1,14 @@
-<div class="form-group" [ngClass]="{'form-group-error': !fixedListControl.valid && fixedListControl.touched}">
+<div class="form-group" [ngClass]="{'form-group-error': false}">
```
mario-paniccia: do we need `[ngClass]="{'form-group-error': false}"` at all now?

kiran-yenigala-hmcts: Removed

src/shared/components/palette/fixed-list/write-fixed-list-field.component.ts
kiran-yenigala-hmcts: ok.. Will ask Ashley to test dynamic lists as well to see this change

src/shared/components/palette/fixed-list/write-fixed-list-field.html
```diff
@@ -1,16 +1,14 @@
-<div class="form-group" [ngClass]="{'form-group-error': !fixedListControl.valid && fixedListControl.touched}">
+<div class="form-group">
 
   <label [for]="id()">
     <span class="form-label" *ngIf="caseField.label">{{caseField | ccdFieldLabel}}</span>
     <span class="form-hint" *ngIf="caseField.hint_text">{{caseField.hint_text}}</span>
-    <span class="error-message" *ngIf="fixedListControl.errors && fixedListControl.touched">{{fixedListControl.errors | ccdFirstError}}</span>
```
mario-paniccia: we have removed `ccdFirstError` pipe. If you look at FirstErrorPipe, this pipe is used to provide common standard error messages. For example if the field is required, the common error message should be:
`'This field is required'`
What error message are we getting now?

mario-paniccia: we need to ask QAing around error messages

PR #493, opened 2020-01-10
RDM-6965 - Jurisdiction-Level Shuttering

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-6965

### Change description ###
Introduce new jurisdiction level configs

```
[ ] Yes
[x] No
```


src/shared/domain/definition/jurisdiction-config.model.ts
```diff
@@ -0,0 +1,4 @@
+export class JurisdictionConfig {
```
MSancaktutar: should we have named this as JurisdictionUIConfig, as we named it so elsewhere? -- to avoid confusion with a more generic sense of jurisdiction configuration by definition files.

danlysiak: Addressed 👍

PR #499, opened 2020-01-21
Bug/eui 1305 date vlidation aa

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-1305

### Change description ###

Applied interim solution where when the current page has errors and the user navigates to previous page, current page fields get their values reset.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -120,7 +120,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
 
   toPreviousPage() {
     let caseEventData: CaseEventData = this.buildCaseEventData();
-    this.updateFormData(caseEventData);
+    this.updateFormData(caseEventData, true);
     this.previous();
   }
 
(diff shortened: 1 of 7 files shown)


src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -144,11 +144,11 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
     }
   }
 
-  updateFormData(jsonData: CaseEventData): void {
+  updateFormData(jsonData: CaseEventData, previous: boolean): void {
```
andrewfolga: Can we call `previous` `isPrevious` to follow boolean convention?

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -173,6 +173,14 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
     }
   }
 
+  resetFormControls(formGroup: FormGroup, caseFieldId: string): void {
```
andrewfolga: Can we add comment here on why we need to do this? Also, this is an intermediate solution as I understand correct?

PR #538, opened 2020-06-08
Eui 1940/case selection

kpokala:

### Change description ###

```
[ ] Yes
[ ] No
```


src/shared/components/search-result/search-result.component.ts
```diff
@@ -109,6 +121,72 @@ export class SearchResultComponent implements OnChanges {
     }
   }
 
+  public clearSelection(): void {
```
adnanakgun, line 124: Actually, can we have some tests for these new sections?

src/shared/components/search-result/search-result.component.ts
```diff
@@ -109,6 +121,72 @@ export class SearchResultComponent implements OnChanges {
...
+      this.resultView.results.forEach(c => {
+        if (!this.isSelected(c)) {
+          this.selectedCases.push(c);
+        }
+      });
+    }
+    this.selection.emit(this.selectedCases);
+  }
+
+  public changeSelection(c: SearchResultViewItem): void {
```
adnanakgun, line 153: Should this function not be doing this.selection.emit(this.selectedCases); as well?

ronaldmansveld: Yes it should

PR #548, opened 2020-07-23
Feature/eui 1753 elastic search aa

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-1753

### Change description ###

Added new method to target ES endpoint

```
[ ] Yes
[x] No
```


src/shared/services/search/search.service.ts
```diff
@@ -39,6 +39,23 @@ export class SearchService {
       );
   }
 
+  public searchCases(caseTypeId: string,
+                metaCriteria: object, caseCriteria: object, view?: SearchView, sort?: string): Observable<{}> {
+    const url = this.appConfig.getCaseDataUrl() + `/internal/searchCases?ctid=${caseTypeId}&usecase=${view}`;
```
udaydenduluri33, line 44: Can you move this url logic to a separate method and write few unit test cases?

PR #585, opened 2020-09-21
EUI-2344: Hide generic callback error message

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-2344

### Change description ###
Hide the generic callback error message if a custom configured one has been provided by a Service. Applies to the Case Edit page, Case Edit Submit page, and Case Viewer component.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/error/callback-errors.html
+++ b/src/shared/components/error/callback-errors.html
@@ -10,6 +10,8 @@
       </li>
     </ul>
   </ng-container>
+  <!-- Add a break for spacing if there are both errors and warnings -->
+  <br *ngIf="hasErrors() && hasWarnings()">
   <ng-container *ngIf="hasWarnings()">
     <h3 class="heading-h3 error-summary-heading">
       Warnings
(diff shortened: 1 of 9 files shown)


src/shared/components/error/callback-errors.html
```diff
@@ -10,6 +10,8 @@ <h3 class="heading-h3 error-summary-heading">
       </li>
     </ul>
   </ng-container>
+  <!-- Add a break for spacing if there are both errors and warnings -->
+  <br *ngIf="hasErrors() && hasWarnings()">
```
udaydenduluri33, line 14: @Daniel-Lam  should this be || or && ?

udaydenduluri33 (approved):
LGTMd

Mo-Lala84 (approved):
approved

src/shared/components/error/callback-errors.html
Daniel-Lam: @udaydenduluri33 the `&&` is intentional because the spacing break is required only if there are both errors and warnings.

PR #608, opened 2020-11-09
EUI-2739: Fix "retain hidden fields" functionality for all field types

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-2739

### Change description ###
Functionality fixed for all field types, including Complex and Collection types.

```
[ ] Yes
[x] No
```


Mo-Lala84 (approved):
approved

rayliangatos (approved):
Approved and tested with the FR case types - passed, but we need a full regression tests from QA on create cases flow for different Jurisdition and caseType.

PR #613, opened 2020-11-26
Bug/eui 2741 performance issue tribunals DO NOT MERGE TO MASTER

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-2741

### Change description ###
Performance issue which impacts Employment Tribunals usage of Manage Cases. Issue is clearly apparent when loading a case with large hearing and judgement collections.

```
[ ] Yes
[x ] No
```


src/shared/components/palette/history/event-log/event-log-table.component.spec.ts
```diff
@@ -108,7 +108,8 @@ describe('EventLogTableComponent', () => {
       expect(firstRowCells.length).toBe(3);
       let firstEvent = EVENTS[0];
 
-      const timeZoneOffset = -(new Date().getTimezoneOffset());
+      const date = new Date(2017, 4, 10); // 10th May, 2017
```
paul-graham, line 111: What is the significance of specifying 10th May 2017 and not using the current date?

paddybasi (approved):

src/shared/components/palette/history/event-log/event-log-table.component.spec.ts
paddybasi: I would presume nothing, it's just a test right?

paulhowes-HMCTS: Its code that was previously there and from what I gather it needs to be a date that falls in the summer otherwise the test will fail (hence the fix).

paul-graham: If you change a test, it's usually because it wasn't passing without the change. I'm just checking that this change isn't simply to get through a failing test without addressing the underlying failure. :)

paul-graham: > Its code that was previously there and from what I gather it needs to be a date that falls in the summer otherwise the test will fail (hence the fix).

Cool cool cool. Then have yourself an approval.

paul-graham (approved):
All looks fine to me.

adnanakgun (approved):

src/shared/components/palette/history/event-log/event-log-table.component.spec.ts
adnanakgun: Actually @paul-graham is correct. I believe this test only fails locally. In fact, I believe it will fail on the server because of this change.

Mo-Lala84 (approved):
approved

PR #618, opened 2020-12-01
EUI-2739: Fix "retain hidden fields" functionality-DONT MERGE TO MASTER UNTIL LIVE RELEASE

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-2739
https://tools.hmcts.net/jira/browse/EUI-3622
https://tools.hmcts.net/jira/browse/EUI-3868
https://tools.hmcts.net/jira/browse/EUI-2744

### Change description ###
Fixes for Complex collection, nested Complex, and nested Complex collection types.

```
[x] Yes
[ ] No
```


rayliangatos (approved):
Approved and this needs to release from feature branch with a feature pre-release tag, once QA tests are passed, and all other teams are happy to go live then merge back to master.

src/shared/services/form/form-value.service.ts
```diff
@@ -218,4 +219,14 @@ export class FormValueService {
     }
   }
 
+  /**
+   * Recursive check of an object and its descendants for the presence of any non-empty values.
+   *
+   * @param object The object to check
+   */
+  private containsNonEmptyValues(object: object): boolean {
```
andywilkinshmcts, line 227: This worries me a bit.
1) It uses some(), so mostly it will be short circuited, but what if there is a circular reference in the object hierarchy in the first key it comes to? Stack overflow, presumably. I'm not sure if it will be possible to engineer a circular ref with CCD definitions, but I'm not willing to rule it out - for example people could do funky things with callbacks. The only thing you could would be to keep a map of all the objects you've already seen, but this would probably reduce performance and increase memory size.
2) It might be better to do a breadth first search rather than depth first. This will improve performance and reduce the impact of 1, since most objects will have a primitive member. As written you recurse into any objects as you come to them. What you could do instead is to iterate over the keys, check for a non-null primitive member, and push all the object references on to a list. If you don't find a non-null primitive then iterate over the object list
 ```
 var iObject = [];
  var ret = values.some(x => 
    (x !== null &&
      (typeof x == 'object'  ? iObject.push (x) : x !== '')
  if (!ret) {
    return iObject.some(y =>  this.containsNonEmptyValues(y)) :
 } else {
   return true
}
```

**Review by andywilkinshmcts — CHANGES_REQUESTED** [review]
Probably OK, but please consider my comment about containsEmptyValues

File: src/shared/services/form/form-value.service.ts (hunk 1 above)
Comment (Daniel-Lam) [reply src/shared/services/form/form-value.service.ts:227]: Hi @andywilkinshmcts - just checking something... isn't it the case that, given `var iObject = []`, then `iObject.push(x)` will always be truthy (because it returns the new length of the `iObject` array) - and thus, `(typeof x == 'object' ? iObject.push(x) : x !== '')` will always evaluate to `true`, if `x` is an object?

In this case, `ret` will never be `false`, which means we would never execute `iObject.some(y =>  this.containsNonEmptyValues(y))`...?

What if every `x` was an empty object, or an object whose keys referenced only `null` values? These would not be tested by the code above.

Comment (andywilkinshmcts) [reply src/shared/services/form/form-value.service.ts:227]: `(typeof x == 'object' ? iObject.push(x) : x !== '')` will always evaluate to `true`, if `x` is an object?
> 
> In this case, `ret` will never be `false`, which means we would never execute `iObject.some(y => this.containsNonEmptyValues(y))`...?
Quite correct, therefore `(typeof x == 'object' ? !iObject.push(x) : x !== '')

So it always fails if we push something

Comment (Daniel-Lam) [reply src/shared/services/form/form-value.service.ts:227]: Hi @andywilkinshmcts - makes sense... I'll add the missing `!` to `iObject.push(x)`.

**Review by andywilkinshmcts — DISMISSED** [review]

File: RELEASE-NOTES.md [hunk 2]
```diff
@@ -1,4 +1,18 @@
 ## RELEASE NOTES
+### Version 3.0.0-epsilon
```
paul-graham, line 2: Will these all eventually end up being simply `Version 3.0.0` without all the pre-release versions mentioned?

src/shared/components/case-editor/case-edit-submit/case-edit-submit.component.ts
```diff
@@ -91,7 +91,11 @@ export class CaseEditSubmitComponent implements OnInit, OnDestroy {
 
   submit(): void {
     this.isSubmitting = true;
-    let caseEventData: CaseEventData = this.formValueService.sanitise(this.editForm.value) as CaseEventData;
+    let caseEventData: CaseEventData = {
```
paul-graham, line 94: Should I `let` you away with this one? I'm not convinced...

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "2.79.4",
+  "version": "3.0.0-epilson",
```
paul-graham, line 3: Shouldn't have the `-epsilon` suffix.

src/shared/components/palette/case-link/write-case-link-field.component.ts
```diff
@@ -28,6 +28,14 @@ export class WriteCaseLinkFieldComponent extends AbstractFieldWriteComponent imp
     }
     this.caseReferenceControl = this.caseLinkGroup.controls['CaseReference'];
     this.caseReferenceControl.setValidators(this.caseReferenceValidator());
+
+    // Ensure that all sub-fields inherit the same value for retain_hidden_value as this parent; although a CaseLink
+    // field uses the Complex type, it is meant to be treated as one field
+    if (this.caseField) {
```
paul-graham, line 34: Should this also check that it's a `Complex` field, considering what it does within the condition?

paul-graham (commented):
A couple of suggestions and questions. Nothing that'll block it... even with the `let` in place.

RELEASE-NOTES.md
Daniel-Lam: Yes, that's correct. All of these Greek-alphabet suffix ones are just pre-releases, which will move to `3.0.0` once QA approves everything for release to Production.

src/shared/services/form/form-value.service.ts
```diff
@@ -218,8 +218,13 @@ export class FormValueService {
...
 
     let sanitisedObject = {};
+    const documentFieldKeys = ['document_url', 'document_binary_url', 'document_filename'];
     Object.keys(rawObject).forEach(key => {
-      if ('CaseReference' === key) {
+      // If the key is one of documentFieldKeys, it means the field is of Document type. If the value of any of these
+      // properties is null, the entire sanitised object to be returned should be null
+      if (documentFieldKeys.indexOf(key) > -1 && rawObject[key] == null) {
+        sanitisedObject = null;
+      } else if ('CaseReference' === key) {
```
rayliangatos, line 227: is there unit test to cover this form-value.service.ts?

rayliangatos (commented):
> ### JIRA link (if applicable)
> https://tools.hmcts.net/jira/browse/EUI-2739
> https://tools.hmcts.net/jira/browse/EUI-3622
> 
> ### Change description
> Fixes for Complex collection, nested Complex, and nested Complex collection types.
> 
> **Does this PR introduce a breaking change?** (check one with "x")
> 
> ```
> [ ] Yes
> [x] No
> ```

Should this be breaking changes?

src/shared/components/case-editor/case-edit-submit/case-edit-submit.component.ts
Daniel-Lam: Haha 😄 - it's a fair catch!

package.json
Daniel-Lam: The suffix will be dropped prior to release to Production. For now, it's there only because I've not created a release yet - just a tag for QA to test.

src/shared/components/palette/case-link/write-case-link-field.component.ts
Daniel-Lam: Hmm - that's a good point. `caseField.field_type` is a mandatory property, so it's worth checking this is equal to `Complex`. There is also the same thing in `write-organisation-field.component.ts`, so I'll add it there as well.

src/shared/services/form/form-value.service.ts
Daniel-Lam: @rayliangatos - there isn't an explicit unit test for this (because the `sanitiseObject` function is `private`). However, it should already be covered by the higher-level functional tests in `case-edit-submit.component.spec.ts`. I'll have a look at adding a unit test if possible.

rayliangatos (approved):

paul-graham (dismissed):
Nice. Thanks for making the requested changes, @Daniel-Lam. Approved. 👍

rayliangatos (dismissed):

paul-graham (dismissed):

rayliangatos (dismissed):

paul-graham (dismissed):
Nothing changed since I approved this so still all good. 👍

rayliangatos (dismissed):

rayliangatos (approved):

src/shared/services/fields/fields.purger.ts
```diff
@@ -66,58 +66,227 @@ export class FieldsPurger {
...
-  private hasShowConditionField(case_field, formFields): boolean {
+  private hasShowConditionField(case_field: CaseField, formFields: any): boolean {
     return case_field.show_condition && formFields.data[this.getShowConditionKey(case_field.show_condition)];
   }
 
-  private getShowConditionKey(show_condition): string {
-    return show_condition.split('=')[0];
+  private getShowConditionKey(show_condition: string): string {
+    // Need to allow for negated conditions, i.e. !=, as well as regular ones (=)
+    return show_condition.split(/!=|=/)[0];
```
paul-graham, line 87: Should this also check for `'CONTAINS'`?

paul-graham (commented):
One quick question...

src/shared/services/fields/fields.purger.ts
Daniel-Lam: @paul-graham - I have now added this

rayliangatos (dismissed):

paul-graham (approved):

paul-graham (dismissed):

rayliangatos (dismissed):

src/shared/services/fields/fields.utils.ts
```diff
@@ -158,15 +158,15 @@ export class FieldsUtils {
     }
   };
 
-  private static getMoneyGBP(fieldValue: any): any {
+  private static getMoneyGBP(fieldValue: any): string {
     return fieldValue ? FieldsUtils.currencyPipe.transform(fieldValue / 100, 'GBP', 'symbol') : fieldValue;
```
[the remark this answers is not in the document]
Daniel-Lam: Ah, I think this may have come in when I merged `Release` into my PR... I _knew_ I shouldn't have done that 😄. Will change it back to `any`.

paul-graham (dismissed):
Yeah, the `getMoneyGBP` thing was my fault. Well, the signature change was anyway - the dodgy logic, I'm not sure... but I'll sort it out anyway. 🤦

src/shared/services/fields/fields.utils.ts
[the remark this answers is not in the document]
Daniel-Lam: Following discussion with @paul-graham, he'll resolve the issue above.

PR #636, opened 2020-12-10
Accessibility - default focus to top of page on load

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-2020

### Change description ###
Default the focus to the top of the page for accessibility using keyboard only and screen reader

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -82,9 +82,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
         }
       }
     });
-    if (document.getElementById('main-content')) {
-      document.getElementById('main-content').focus();
-    }
+    this.setFocusToTop();
   }
 
   ngAfterViewChecked(): void {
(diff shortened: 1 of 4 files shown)


adnanakgun (approved):

udaydenduluri33 (approved):
LGTMp

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -306,4 +303,11 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
     }
     return caseEventData;
   }
+
+  private setFocusToTop() {
+    const topContainer = document.getElementById('top');
```
adnanakgun, line 308: Toolkit should not be dependant on the consumer implementing ID's, however we seem to have done this already in the past.

paulhowes-HMCTS: Agreed - I am aiming to add functionality which removes the dependency on the consumer implementing IDs, so this is only an intermediary change.

PR #657, opened 2021-01-20
Feature/EUI 2029 wcag status messages phillip whitaker

phillip-whitaker:

https://tools.hmcts.net/jira/browse/EUI-2029

### Change description ###

The following changes resolve an issue with item 2 & 3 in EUI-2029. Namely, making sure that after a search the focus of the page is taken to the search result.

```
[ ] Yes
[X] No
```


src/shared/components/case-editor/case-edit-page/case-edit-page.html
```diff
@@ -13,7 +13,8 @@ <h2 *ngIf="getCaseId()" class="heading-h2">
 
 <!-- Generic error heading and error message to be displayed only if there are no specific callback errors or warnings, or no error details -->
 <div *ngIf="error && !(error.callbackErrors || error.callbackWarnings || error.details)" class="error-summary"
-     role="group" aria-labelledby="edit-case-event_error-summary-heading" tabindex="-1">
+     role="status" aria-labelledby="edit-case-event_error-summary-heading" tabindex="-1">
```
Daniel-Lam, line 16: @phillip-whitaker - just a quick one... what's the implication of changing the `role` from `group` to `status`?

src/shared/components/search-result/search-result.component.html
```diff
@@ -1,53 +1,53 @@
 <table *ngIf="hasResults() || hasDrafts()">
   <caption>
-    <h2 class="heading-h2">{{ caseState ? 'Case List' : 'Search result' }}</h2>
+    <h2 class="heading-h2" id="search-result-heading__text" tabindex="-1">{{ caseState ? 'Case List' : 'Search result' }}</h2>
     <div *ngIf="(hasResults() || hasDrafts())" class="pagination-top"
-         attr.aria-label="{{ getTotalResults() }} results have been found">
-      <span class="text-16">Displaying {{ getFirstResult() }} - {{ getLastResult() }} out of {{ getTotalResults() }} results</span>
+      attr.aria-label="{{ getTotalResults() }} results have been found" role="status">
```
Daniel-Lam, line 5: @phillip-whitaker - what happens if `getTotalResults()` equals 1? Is it OK to have `1 results have been found`?

src/shared/components/search-result/search-result.component.html
```diff
@@ -1,53 +1,53 @@
...
+      <span class="text-16" id="search-result-summary__text">Displaying {{ getFirstResult() }} - {{ getLastResult() }} 
+        out of {{ getTotalResults() }} results</span>
```
Daniel-Lam, line 7: Same comment as above for `getTotalResults()`.

src/shared/components/workbasket-filters/workbasket-filters.component.ts
```diff
@@ -279,4 +280,13 @@ export class WorkbasketFiltersComponent implements OnInit {
     this.selected.caseType = undefined; // option should be blank rather than "Select a value" in case of reset.
     this.selectedJurisdictionCaseTypes = null;
   }
+
+  private setFocusToTop() {
```
Daniel-Lam, line 284: @phillip-whitaker - more of a refactoring optimisation, is there anywhere you can put the `setFocusToTop()` function, which can be used by both components? This was also defined in `search-filters.component.ts`.

Daniel-Lam (changes requested):
@phillip-whitaker - just a few minor things to look at, but the rest is OK.

src/shared/components/search-result/search-result.component.html
phillip-whitaker: I think so, I don't really want to complicate it.

src/shared/components/case-editor/case-edit-page/case-edit-page.html
phillip-whitaker: I am not hugely sure, other than it is what the WCAG report suggested we did in the report. I think it changes the importance for screen readers essentially.

Daniel-Lam: OK, no worries - thanks.

src/shared/components/search-result/search-result.component.html
Daniel-Lam: Ah, that's fair enough - caught my eye, so just thought I'd check.

src/shared/components/search-result/search-result.component.html
Daniel-Lam: Has been explained already by @phillip-whitaker in the previous comment.

src/shared/components/workbasket-filters/workbasket-filters.component.ts
phillip-whitaker: I agree about the scrollTo function etc. It would be nice in a common function. However.. can I be cheeky and ask they we don't? Purely because it is an old change and would just like to get it through without revisiting everything it did etc.

Daniel-Lam: @phillip-whitaker - fair enough, it's a small function so there probably isn't that much to be gained. Maybe it could be refactored in future...

Daniel-Lam (approved):

PR #659, opened 2021-01-25
Bug/EUI 2830 dynamic lists in collections

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-2830

### Change description ###

Implemented multiple recursive functions and workarounds for the json

```
[ ] Yes
[x] No
```


src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -104,31 +105,65 @@ export class CasesService {
...
+            };
+            field.formatted_value = {
+              ...field.formatted_value,
+              ...field.value
+            };
+          } else {
+            this.setDynamicListDefinition(field, field.field_type, rootCaseField);
+          }
+        } catch (error) {
+          console.log(error);
```
udaydenduluri33, line 142: Probably you could remove this console.log

src/shared/services/form/field-type-sanitiser.ts
```diff
@@ -13,33 +17,48 @@ export class FieldTypeSanitiser {
...
+          this.sanitiseLists(caseField.field_type.complex_fields, editForm[caseField.id]);
+          break;
+
+        case FieldTypeSanitiser.FIELD_TYPE_COLLECTION:
+          editForm[caseField.id].forEach(formElement => {
+            this.sanitiseLists(caseField.field_type.collection_field_type.complex_fields, formElement.value);
+          });
+          break;
+
+        default:
```
udaydenduluri33, line 39: Can we simply return a value instead of doing nothing for default case?

src/shared/services/form/form-value.service.ts
```diff
@@ -161,7 +161,7 @@ export class FormValueService {
   }
 
   sanitiseDynamicLists(caseFields: CaseField[], editForm: any): any {
-    return this.fieldTypeSanitiser.sanitiseLists(caseFields, editForm);
+    return this.fieldTypeSanitiser.sanitiseLists(caseFields, editForm['data']);
```
udaydenduluri33, line 164: Is `editForm.data` better instead of `editForm['data'] `

udaydenduluri33 (approved):
LGTMa

src/shared/components/case-editor/services/cases.service.ts
adnanakgun: This handles the error of the try-catch, so I think we can leave it for now. As this is highly recursive code I want to be able to tell if the app runs into "Maximum call stack size exceeded" issues and rethink the solution in the future.

src/shared/services/form/field-type-sanitiser.ts
adnanakgun: Function doesn't return anything actually. Do you think it would be OK if I remove the default completely?

src/shared/services/form/form-value.service.ts
adnanakgun: Done

Mo-Lala84 (approved):
lgtm

PR #664, opened 2021-01-28
work allocation task service is now feature toggled

udaydenduluri33:
This PR helps to feature toggle the Work Allocation services which are tied up in Case Service

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/shared/components/case-editor/services/cases.service.ts
+++ b/src/shared/components/case-editor/services/cases.service.ts
@@ -297,13 +297,16 @@ export class CasesService {
   }
 
   private processTasksOnSuccess(caseData: any, eventData: any): void {
-    this.workAllocationService.completeAppropriateTask(caseData.id, eventData.id)
-      .subscribe(() => {
-        // Success. Do nothing.
-      }, error => {
-        // Show an appropriate warning about something that went wrong.
-        console.warn('Could not process tasks for this case event', error);
-      });
+    // This is used a feature toggle to 
+    // control the work allocation
+    if(this.appConfig.getWorkAllocationApiUrl()) {
+        this.workAllocationService.completeAppropriateTask(caseData.id, eventData.id)
+          .subscribe(() => {
+            // Success. Do nothing.
+          }, error => {
+            // Show an appropriate warning about something that went wrong.
+            console.warn('Could not process tasks for this case event', error);
+          });
+    }
   }
-
 }
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,5 @@
 ## RELEASE NOTES
-### 2.68.0-feature-toggle-work-allocation
+### 2.67.10-feature-toggle-work-allocation
 **Feature Toggle Work Allocation** Hot Fix
 
 ### Version 2.67.9-ie-textbox-expand-issue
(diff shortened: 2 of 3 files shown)


andywilkinshmcts (changes requested):
Please change the version number. I am using 2.68.xxx for the perf fix branch. Please use either 2.67.10 or 2.69.0

src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -297,13 +297,16 @@ export class CasesService {
...
-        // Show an appropriate warning about something that went wrong.
-        console.warn('Could not process tasks for this case event', error);
-      });
+    // This is used a feature toggle to 
+    // control the work allocation
+    if(this.appConfig.getWorkAllocationApiUrl()) {
+        this.workAllocationService.completeAppropriateTask(caseData.id, eventData.id)
+          .subscribe(() => {
+            // Success. Do nothing.
+          }, error => {
```
phillip-whitaker, line 306: Do we need the comments?

src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -297,13 +297,16 @@ export class CasesService {
...
-    this.workAllocationService.completeAppropriateTask(caseData.id, eventData.id)
-      .subscribe(() => {
-        // Success. Do nothing.
-      }, error => {
-        // Show an appropriate warning about something that went wrong.
-        console.warn('Could not process tasks for this case event', error);
-      });
+    // This is used a feature toggle to 
+    // control the work allocation
+    if(this.appConfig.getWorkAllocationApiUrl()) {
```
phillip-whitaker, line 302: could this be refactored to make the comment redundant...

```
const workAllocationFeatureToggle = this.appConfig.getWorkAllocationApiUrl()

if (workAllocationFeatureToggle ) {...
```

udaydenduluri33:
> 2.67.10

Done now. Thanks

src/shared/components/case-editor/services/cases.service.ts
udaydenduluri33: @phillip-whitaker Agree but, I want the comment to make it explicit to the developer that config needs a null which is used a feature toggle.

src/shared/components/case-editor/services/cases.service.ts
udaydenduluri33: @phillip-whitaker Agree but, I want the comment to make it explicit to the developer that config needs a null which is used a feature toggle.

andywilkinshmcts (approved):

src/shared/components/case-editor/services/cases.service.ts
paul-graham: @phillip-whitaker, whilst I don't disagree with putting it in a `const` first, it shouldn't have a name that implies that it's a feature toggle in and of itself, because it is indeed the URL to use.

The mechanism @udaydenduluri33 has put in place here is to avoid the need for _actual_ feature toggles in the CCD Case UI Toolkit and I think a comment explaining why this one behaves differently to the others is probably sensible. :)

paul-graham (approved):

Mo-Lala84 (approved):
approved

PR #666, opened 2021-01-28
E UI 3232 alert service edition

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-3232

### Change description ###
Changing alert service messages:
- Allowing messages of different types to be shown at once
- Adding correct styling

```
[ x ] Yes
[ ] No
```

### DIFF
--- a/src/components/banners/alert/alert.component.html
+++ b/src/components/banners/alert/alert.component.html
@@ -1,5 +1,5 @@
 <div [ngClass]="{'hmcts-banner hmcts-banner--warning':type === alertMessageType.WARNING || type === alertMessageType.ERROR}">
-  <div [ngClass]="{'hmcts-banner hmcts-banner--success':type === alertMessageType.SUCCESS">
+  <div [ngClass]="{'hmcts-banner hmcts-banner--success':type === alertMessageType.SUCCESS}">
     <ng-container [ngSwitch]="type">
       <ng-container *ngSwitchCase="alertMessageType.WARNING">
         <svg class="hmcts-banner__icon" fill="currentColor" role="presentation" focusable="false"
--- a/src/components/banners/alert/alert.component.html
+++ b/src/components/banners/alert/alert.component.html
@@ -1,7 +1,35 @@
-<div class="alert alert-{{type}}" role="alert">
+<!-- <div class="alert alert-{{type}}" role="alert">
   <div class="alert-container">
     <i class="icon {{type | cutAlertIconClass}}" role="presentation" [attr.aria-hidden]="!showIcon" *ngIf="showIcon"></i>
     <div class="alert-message" data-ng-transclude>
+      <ng-content></ng-content>
+    </div>
+  </div>
+</div> -->
+<div [ngClass]="{'hmcts-banner hmcts-banner--warning':type === alertMessageType.WARNING || type === alertMessageType.ERROR}">
+  <div [ngClass]="{'hmcts-banner hmcts-banner--success':type === 'success'}">
+    <ng-container [ngSwitch]="type">
+      <ng-container *ngSwitchCase="alertMessageType.WARNING">
+        <svg class="hmcts-banner__icon" fill="currentColor" role="presentation" focusable="false"
+          xmlns="http://www.w3.org/2000/svg" viewBox="0 0 25 25" height="25" width="25">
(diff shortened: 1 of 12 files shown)


udaydenduluri33 (approved):
LGTMc

Mo-Lala84 (approved):
approved

src/components/banners/alert/alert.component.html
```diff
@@ -1,8 +1,28 @@
-<div class="alert alert-{{type}}" role="alert">
-  <div class="alert-container">
-    <i class="icon {{type | cutAlertIconClass}}" role="presentation" [attr.aria-hidden]="!showIcon" *ngIf="showIcon"></i>
-    <div class="alert-message" data-ng-transclude>
+<div [ngClass]="{'hmcts-banner hmcts-banner--warning':type === alertMessageType.WARNING || type === alertMessageType.ERROR}">
+  <div [ngClass]="{'hmcts-banner hmcts-banner--success':type === 'success'}">
```
paul-graham, line 2: Why do we have have an outer div with a warning class and an inner div with a success class?

paul-graham: Also, why are WARNING and ERROR using constants but success is just a string?

src/components/banners/alert/alert.component.html
```diff
@@ -1,8 +1,28 @@
...
+        </svg>
+      </ng-container>
+    </ng-container>
+    <div class="hmcts-banner__message">
+      <span class="hmcts-banner__assistive">{{type}}</span>
       <ng-content></ng-content>
     </div>
   </div>
-</div>
+</div>
```
paul-graham, line 28: Carriage return needed.

src/components/banners/alert/alert.component.ts
```diff
@@ -12,9 +18,11 @@ export class AlertComponent {
   public static readonly TYPE_WARNING = 'warning';
   public static readonly TYPE_CONFIRMATION = 'confirmation';
   public static readonly TYPE_SUCCESS = 'success';
+  public static readonly TYPE_ERROR = 'error';
 
   @Input()
-  public type: 'warning' | 'confirmation' | 'success' = AlertComponent.TYPE_WARNING;
+  public type: AlertMessageType;
```
paul-graham, line 24: Has the 'confirmation' type disappeared? Was that intentional?

src/shared/services/alert/alert.service.ts
```diff
@@ -8,70 +8,116 @@ import { Alert } from '../../domain/alert/alert.model';
...
   isPreserveAlerts(): boolean {
     return this.preserveAlerts;
   }
 
-  message(message: string): void {
-    this.push({
-      level: 'message',
-      message: message
-    });
+  preserveMessages(message: string) {
```
paul-graham, line 115: This should have a return type of `string`.

paul-graham (changes requested):

src/components/banners/alert/alert.component.html
connorpgpmcelroy: To answer the first question, I couldn't get all the styles set on one div. I think because the subscription of the observers goes through this code three times the class would be the same for all of them if we wanted to put it in one div (even with ngClass). I attempted to get the styling correct for all messages with more clarity but this was the simplest way I could find that displayed the desired styling.

connorpgpmcelroy: Don't know why I've left success as a string though, I'll change that

src/components/banners/alert/alert.component.ts
connorpgpmcelroy: I'll get rid of this. The types were already in there so it should be dead code now really

connorpgpmcelroy: Need confirmation on what to do about alert-icon-class-pipe.ts regarding this

src/components/banners/alert/alert.component.html
connorpgpmcelroy: What are you thoughts on my answer with regard to the first question? I could replicate the issue I was getting with ngClass and demo it to you to indicate the issues I was having with regards to that styling?

paul-graham (approved):

PR #674, opened 2021-02-03
E UI 3229 task completion iac role

udaydenduluri33:
https://tools.hmcts.net/jira/browse/EUI-3229
(which follows on from related changes in https://tools.hmcts.net/jira/browse/EUI-3232)

### Change description ###
Improving task event completion

```
[] Yes
[ x ] No
```

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,8 +1,7 @@
 ## RELEASE NOTES
 
-### Version 2.72.0-alert-service-update
+### Version 2.72.0-task-event-completion
 **EUI-3229** Task event completion - part 2
-**EUI-3232** CCD UI toolkit Alert service message improvements
 
 ### Version 2.71.4-optional-top-level-documents
 Fix EUI-3403 - fix for optional top-level documents incorrectly showing.
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,6 @@
 ## RELEASE NOTES
 
-### Version 2.70.1-alert-service-update
+### Version 2.71.0-alert-service-update
 **EUI-3229** Task event completion - part 2
 **EUI-3232** CCD UI toolkit Alert service message improvements
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,4 +1,9 @@
 ## RELEASE NOTES
+
+### Version 2.69.2-alert-service-update
+**EUI-3229** Task event completion - part 2
+**EUI-3232** CCD UI toolkit Alert service message improvements
+
 ### Version 2.69.1-fix-replace-document-issue
 **EUI-2243** Fix upload replacement document issue
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,6 @@
 ## RELEASE NOTES
 
-### Version 3.0.0-alert-service-update
+### Version 2.71.0-alert-service-update
 **EUI-3229** Task event completion - part 2
 **EUI-3232** CCD UI toolkit Alert service message improvements
 
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,6 +1,6 @@
 ## RELEASE NOTES
 
-### Version 2.71.0-alert-service-update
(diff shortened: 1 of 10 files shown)


RELEASE-NOTES.md
```diff
@@ -1,6 +1,12 @@
 ## RELEASE NOTES
+
+### Version 2.70.1-alert-service-update
```
paul-graham, line 3: As a minimum, this is a minor release, not a patch. It should go up to `2.71.0`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "2.70.0-fix-alert-service",
+  "version": "2.70.1-alert-service-update",
```
paul-graham, line 3: As a minimum, this is a minor release, not a patch. It should go up to `2.71.0`.

src/shared/components/case-editor/services/work-allocation.service.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { Response, ResponseOptions } from '@angular/http';
-import { Observable, throwError } from 'rxjs';
+import { Observable, of, throwError } from 'rxjs';
```
paul-graham, line 2: I'm not seeing where you're using `of` in here. You're always using `Observable.of`, which makes the `of` import redundant.

RELEASE-NOTES.md
paul-graham: Also, it doesn't look like this is to do with the alert service...?

paul-graham (changes requested):

paul-graham:
The description is also entirely untouched.

RELEASE-NOTES.md
connorpgpmcelroy: Agree with first comment. Regarding second comment, even though the changes are within workallocation service, the changes have been made to the alert service and its ability to display messages in certain circumstances. Plus it's a reference to the earlier ccd-ui-toolkit change made

package.json
connorpgpmcelroy: Agreed

src/shared/components/case-editor/services/work-allocation.service.spec.ts
connorpgpmcelroy: Agreed

RELEASE-NOTES.md
paul-graham: I don't see any changes in this release to the alert service. I see it being _used_ by the WA stuff, but no changes to the alert service itself...?

Besides, the name should describe what the primary purpose of this release is for and that looks to be task event completion - the PR is even named for that JIRA. :)

paul-graham:
Hang on, what's the breaking change in here?

RELEASE-NOTES.md
paul-graham: Scrub that, this includes a breaking change. It should be version `3.0.0`.

package.json
paul-graham: Scrub that, this includes a breaking change. It should be version `3.0.0`.

paul-graham (changes requested):

connorpgpmcelroy:
Changing to 3.0.0 as adding in breaking change

src/app.config.ts
```diff
@@ -25,6 +25,9 @@ export abstract class AbstractAppConfig {
   abstract getPrdUrl(): string;
   abstract getCacheTimeOut(): number;
   abstract getWorkAllocationApiUrl(): string;
+  getUserInfoApiUrl(): string {
+    return undefined;
+  }
```
udaydenduluri33, line 30: Do you mind adding a comment why its undefined?

connorpgpmcelroy: It is undefined so it does not cause breaking change on master webapp so that when it is pushed prior to workallocation-6 there are no issues

kuda1992 (approved):
LGTM

src/app.config.ts
paul-graham: @udaydenduluri33, this is specifically not an `abstract` method so it doesn't count as a breaking change.

paul-graham (approved):

PR #681, opened 2021-02-05
added focus link styling for table row anchor links

phillip-whitaker:

https://tools.hmcts.net/jira/browse/EUI-2031

by accident, it also resolves: https://tools.hmcts.net/jira/browse/EUI-2033

### Change description ###

Changed link text so it matches GDS style and satisfies WCAG AA

image

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/search-result/search-result.component.scss
+++ b/src/shared/components/search-result/search-result.component.scss
@@ -6,7 +6,6 @@
 $blue_table: #005ea5;
 $blue_pagination: #005da6;
 $black: #231F20;
-$focus-yellow: #FFDD00;
 
 table {
 
@@ -20,16 +19,6 @@ table {
 
     a {
       float: left;
-      color: $blue_table;
-      background-color: $white;
-
-      &.govuk-link:focus {
-        outline: 3px solid transparent;
-        color: $black;
-        background-color: $focus-yellow;
-        box-shadow: 0 -2px $focus-yellow, 0 4px $black;
-        text-decoration: none;
-      }
     }
   }
 
--- a/src/shared/components/search-result/search-result.component.scss
+++ b/src/shared/components/search-result/search-result.component.scss
@@ -6,6 +6,7 @@
 $blue_table: #005ea5;
 $blue_pagination: #005da6;
 $black: #231F20;
+$focus-yellow: #FFDD00;
 
 table {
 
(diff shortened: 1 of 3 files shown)


src/shared/components/search-result/search-result.component.scss
```diff
@@ -6,6 +6,7 @@
 $blue_table: #005ea5;
 $blue_pagination: #005da6;
 $black: #231F20;
+$focus-yellow: #FFDD00;
```
paul-graham, line 9: Why not use `$govuk-focus-colour` as it's already part of the GDS colour palette?

src/shared/components/search-result/search-result.component.scss
```diff
@@ -16,9 +17,19 @@ table {
   tbody tr td {
     font-size: 16px;
     word-wrap: break-word;
+
     a {
+      float: left;
       color: $blue_table;
       background-color: $white;
+
+      &.govuk-link:focus {
```
paul-graham, line 26: What are we doing to links that means the `govuk-link` class is being overridden in such a way that this focus state isn't working? We shouldn't need to have to re-override it...

phillip-whitaker: The specificity of the current styling here is greater than the generic govuk-link class. Therefore the links do style as expected. Which is why I have added this clause here

src/shared/components/search-result/search-result.component.scss
phillip-whitaker: This is what I talked about in the tech forum. The colour that comes through on govuk-focus-colour is an 'orangey' yellow and does not match the rest of the website if I do so.

paul-graham: Oh, according to the current GDS style guide, that's the same hex colour. We clearly have an out-of-date one somewhere!

src/shared/components/search-result/search-result.component.scss
paul-graham: So I see... but why are we even setting the colour and the background on links within the table? Is that a _different_ blue to the normal link blue? If so, that seems very odd.

paul-graham (approved):

adnanakgun (approved):

PR #684, opened 2021-02-08
EUI-3045: Merged into performance improvements code

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3045

### Change description ###
@phillip-whitaker created a specific dynamic read/write dropdown field. The new write dropdown component checks whether a value property has already been set from the API and if not, sets the value to null so that the default "-- Select an option --" is shown rather than a blank dropdown. This has now been merged into the performance improvements branch with the requisite changes to retain the functionality.

```
[ ] Yes
[x] No
```


phillip-whitaker (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "2.71.0-conditional-show-perf",
+  "version": "2.71.4-optional-top-level-documents",
```
paul-graham, line 3: The latest on `master` is `2.72.1-something-or-other`.

src/shared/components/palette/dynamic-list/read-dynamic-list-field.component.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+    it('render null value as empty string', () => {
+      component.caseField.value = null;
+      fixture.detectChanges();
+
+      expect(de.nativeElement.textContent).toEqual(EMPTY);
+    });
+  });
+
+  describe('ReadDynamicListFieldComponent for dynamiclist type', () => {
+    const VALUE_DYNAMIC_LIST = '{\n' +
```
paul-graham, line 112: A string literal would be much easier to read and avoid the need for the `\n`s entirely.

paul-graham (commented):
Apparently, I can't request changes on my own PR (makes sense) so I'm just commenting. :)

demo/src/app/app.config.ts
```diff
@@ -139,4 +139,8 @@ export class AppConfig extends AbstractAppConfig {
   public getWorkAllocationApiUrl(): string {
     return this.config.work_allocation_api_url;
   }
+
+  public getUserInfoApiUrl(): string {
```
paul-graham, line 143: Really not understanding why this is an addition... it's already on `master`, isn't it?

paul-graham:
Replaced by https://github.com/hmcts/ccd-case-ui-toolkit/pull/697.

PR #685, opened 2021-02-09
Update alert.service.ts

connorpgpmcelroy:
Not directly applicable as fix to resolve braking change but I will link relevant ticket https://tools.hmcts.net/jira/browse/EUI-3229 

### Change description ###
Editing alert service to ensure MC can run correctly

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/shared/services/alert/alert.service.ts
+++ b/src/shared/services/alert/alert.service.ts
@@ -139,7 +139,6 @@ export class AlertService {
 
   // TODO: Remove
   push(msgObject) {
-    console.log('Hi' + msgObject);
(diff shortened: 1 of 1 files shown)


udaydenduluri33 (approved):
LGTMc
I understand this as an interim solution for backfilling your breaking changes

paul-graham (approved):
Rather than remove the bits you've just added in, why not give the `Alert` class an additional property that allows you to differentiate between successes, errors, and warnings, rather than having to observe three different things?

For now, though, approved.

Mo-Lala84 (approved):
approved

PR #687, opened 2021-02-10
EUI-3375: Clear out labels nested within complex types within collections

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3375

### Change description ###
Reintroduced the recursion within collections to clear out unwanted fields within the `value` part of the item. This had been removed due to EUI-3284 but we still need to remove nested labels because the API doesn't like them.

EUI-3284 still failed with just that change so I've added in a check for the display_context being set to 'HIDDEN' when considering whether to delete a hidden field's data at the submission stage. When the display_context is set to 'HIDDEN', it means that it's been overridden by the page wizard to not be seen by the user, rather than having been hidden by the normal show/hide logic.

There are instances where sending hidden data causes the API to return an error - i.e., data has been submitted that shouldn't have been - and others where we need to send it because the API needs them. The distinguishing feature seems to be with `display_context = 'HIDDEN'`, hence this new check.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/services/form/form-value.service.ts
+++ b/src/shared/services/form/form-value.service.ts
@@ -312,6 +312,7 @@ export class FormValueService {
    * @param data The object to be tidied up.
    * @param caseFields The CaseFields that need to be cleaned up.
    * @param clearEmpty Whether or not we should clear out empty, optional, complex objects.
+   * @param clearNonCase Whether or not we should clear out non-case fields at the top level.
    */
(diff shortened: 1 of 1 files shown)


andywilkinshmcts (approved):
Looks like it will work

andywilkinshmcts:
Fixing EUI-3375

PR #692, opened 2021-02-12
EUI-3038: Fix for superfluous null value submission

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-3038

### Change description ###
Ensure that any hidden fields which do not have a value are not given a `null` value on form submission. Furthermore, any hidden fields with `retain_hidden_value` set to `true` are removed from the form submission, since these fields' values should not be updated. This fix applies to all field types, including Complex and Collection types, and combinations thereof.

```
[x] Yes - potentially affects the "show condition" rework by @andywilkinshmcts and @paul-graham
[ ] No
```


src/shared/components/case-editor/case-edit-submit/case-edit-submit.component.ts
```diff
@@ -109,30 +113,118 @@ export class CaseEditSubmitComponent implements OnInit, OnDestroy {
...
+                }
+              }
+            }
+
+            // Discard null, undefined, or empty collections, to avoid submitting a superfluous update for such fields
+            if (caseField.value == null || !FieldsUtils.containsNonEmptyValues(caseField.value)) {
+              delete rawFormValueData[key];
+            }
+            break;
+          default:
```
paul-graham, line 219: This _might_ need a specific case for `DynamicList`s simply because of the shenanigans with the `value` potentially being either an object or a string - if it's an object, it could still not have an _actual_ value defined. That's partly what @phillip-whitaker's recent change addresses.

paul-graham (commented):
Everything in here looks perfectly reasonably to go into the main RHF branch. Just have that one comment regarding `DynamicList` fields.

src/shared/components/case-editor/case-edit-submit/case-edit-submit.component.ts
Daniel-Lam: To be revisited under [EUI-3528](https://tools.hmcts.net/jira/browse/EUI-3528).

PR #693, opened 2021-02-15
EUI-3403: Fix for optional top level documents

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3403

### Change description ###
Document fields were not being registered when there was no value and the field itself was not mandatory, which meant that `show_conditions` would not be evaluated. Cleaned up the logic and the field is now always registered.

Also fixed a lint error in this PR.

```
[ ] Yes
[x] No
```

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,8 @@
 ## RELEASE NOTES
 
+### Version 2.71.4-optional-top-level-documents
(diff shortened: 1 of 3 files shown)


Daniel-Lam (approved):
Approved, @paul-graham - changes look OK to me

andywilkinshmcts (approved):

PR #696, opened 2021-02-18
Replace the button with Anchor element as per GDS

RiteshHMCTS:
https://tools.hmcts.net/jira/browse/EUI-2034

### Change description ###
Replace button with anchor tag

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/create-case-filters/create-case-filters.component.html
+++ b/src/shared/components/create-case-filters/create-case-filters.component.html
@@ -1,4 +1,4 @@
-<form  (ngSubmit)="apply()">
+<form>
   <div class="form-group">
     <label class="form-label" for="cc-jurisdiction">Jurisdiction</label>
     <select class="form-control ccd-dropdown" id="cc-jurisdiction" name="jurisdiction" [formControl]="filterJurisdictionControl" (change)="onJurisdictionIdChange()">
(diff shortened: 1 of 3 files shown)


src/shared/components/create-case-filters/create-case-filters.component.ts
```diff
@@ -9,7 +9,12 @@ import { DefinitionsService, OrderService } from '../../services';
 
 @Component({
   selector: 'ccd-create-case-filters',
-  templateUrl: './create-case-filters.component.html'
+  templateUrl: './create-case-filters.component.html',
+  styles: [`a.disabled {
```
paul-graham, line 13: We don't have anywhere else in the codebase that we use this pattern for styles - the mechanism everywhere else looks like this:
`styleUrls: ['./create-case-filters.component.scss']`

src/shared/components/create-case-filters/create-case-filters.component.html
```diff
@@ -23,5 +23,10 @@
     </select>
   </div>
 
-  <button type="submit" class="button" [disabled]="!isCreatable()">{{startButtonText}}</button>
+  <a (click)="apply()" (keydown.enter)="apply()" [class.disabled]="!isCreatable()" tabindex="{{isCreatable() ? 0 : -1}}" role="button" draggable="false" class="govuk-button govuk-!-margin-top-2 govuk-!-margin-bottom-8 govuk-button--start" data-module="govuk-button">
```
paul-graham, line 26: We've got bindings against functions here. It was agreed in the last tech forum that we should avoid this wherever possible. Can you please refactor the `[class.disabled]` binding so it's bound to a property, not a function.

paul-graham (changes requested):

paul-graham:
There's also no change to `package.json` or `RELEASE-NOTES.md`.

PR #700, opened 2021-02-19
EUI-3045: Dynamic list handle null value

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3045

### Change description ###
Fixed an issue where nulls were not being properly handled for Dynamic List values.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/palette/base-field/field-read.component.ts
+++ b/src/shared/components/palette/base-field/field-read.component.ts
@@ -5,6 +5,8 @@ import { CaseField } from '../../../domain/definition/case-field.model';
 import { FormGroup } from '@angular/forms';
 import { plainToClassFromExist } from 'class-transformer';
 
+const FIX_CASEFIELD_FOR = [ 'FixedList', 'DynamicList' ];
+
 @Component({
   selector: 'ccd-field-read',
   templateUrl: './field-read.html'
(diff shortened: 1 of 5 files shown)


src/shared/components/palette/base-field/field-read.component.ts
```diff
@@ -40,7 +42,7 @@ export class FieldReadComponent extends AbstractFieldReadComponent implements On
       // TODO AW 30/12/20 figure out why FixedLists need plainToClassFromExist
       // Added a check to make sure it's NOT already a CaseField and then
       // assigning it back to this.caseField so we don't create separation.
-      if (this.caseField.field_type.type === 'FixedList' && !(this.caseField instanceof CaseField)) {
+      if (FIX_CASEFIELD_FOR.indexOf(this.caseField.field_type.type) > -1 && !(this.caseField instanceof CaseField)) {
```
phillip-whitaker, line 45: Is the rationale for not using .includes() because of IE support?

phillip-whitaker (approved):

src/shared/components/palette/base-field/field-read.component.ts
paul-graham: Yes, I believe so. We've had to do that same thing elsewhere in CCD Case UI Toolkit.

adnanakgun (approved):

PR #705, opened 2021-02-23
EUI-3490: Case list filter broken

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3490

### Change description ###
The previous implementation used `URLSearchParams`, which automatically ignored `null` properties. Switching to `HttpParams` introduced an issue where such `null`s are not ignored, which had the consequence of sending `"&state=null"` in the query string when the user selected "Any" in the State filter.

Rather than revert back to the deprecated `URLSearchParams` or refactor additional areas for the now built-in `URLSearchParams`, I'm just `null`-checking and making sure they don't get into the query string.

```
[ ] Yes
[x] No
```

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,4 +1,8 @@
 ## RELEASE NOTES
+
+### Version 2.72.6-fix-filter-any-state
+**EUI-3490** Fixed an issue with the Case list filter where "Any" state is selected.
+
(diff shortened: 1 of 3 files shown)


Daniel-Lam (approved):
Approved, @paul-graham - looks good

adnanakgun (approved):

PR #707, opened 2021-02-24
EUI-3494, EUI-3495: Case creation broken

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3494
https://tools.hmcts.net/jira/browse/EUI-3495

### Change description ###
Fix for case creation that was broken when moving from `Http` to `HttpClient`. This was also a problem for the completion of events on the case.

In the midst of this, tidied up some additional points where changes had been made but code had been left in that was redundant (e.g., `map(response => response)` is a dead operation) or confusing (in the same example, it wasn't actually the _response_, it was the response's _body_.)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/services/cases.service.ts
+++ b/src/shared/components/case-editor/services/cases.service.ts
@@ -69,7 +69,6 @@ export class CasesService {
     return this.http
       .get(url)
       .pipe(
-        map(response => response),
         catchError(error => {
           this.errorService.setError(error);
           return throwError(error);
--- a/src/shared/components/case-editor/services/cases.service.ts
+++ b/src/shared/components/case-editor/services/cases.service.ts
@@ -87,7 +87,6 @@ export class CasesService {
     return this.http
       .get(url, {headers, observe: 'body'})
       .pipe(
-        map(response => response),
         catchError(error => {
           this.errorService.setError(error);
           return throwError(error);
@@ -103,19 +102,19 @@ export class CasesService {
    *
    * EUI-2530 Dynamic Lists for Elements in a Complex Type
    *
-   * @param jsonResponse - {}
+   * @param jsonBody - {}
    */
-  private handleNestedDynamicLists(jsonResponse) {
+  private handleNestedDynamicLists(jsonBody: any): any {
 
-    if (jsonResponse.case_fields) {
-      jsonResponse.case_fields.forEach(caseField => {
+    if (jsonBody.case_fields) {
+      jsonBody.case_fields.forEach(caseField => {
         if (caseField.field_type) {
           this.setDynamicListDefinition(caseField, caseField.field_type, caseField);
         }
       });
     }
 
-    return jsonResponse;
+    return jsonBody;
   }
 
   private setDynamicListDefinition(caseField, caseFieldType, rootCaseField) {
@@ -172,7 +171,7 @@ export class CasesService {
                   ignoreWarning?: string): Observable<CaseEventTrigger> {
     ignoreWarning = undefined !== ignoreWarning ? ignoreWarning : 'false';
 
-    let url = this.buildEventTriggerUrl(caseTypeId, eventTriggerId, caseId, ignoreWarning);
+    const url = this.buildEventTriggerUrl(caseTypeId, eventTriggerId, caseId, ignoreWarning);
 
     let headers = new HttpHeaders()
     headers = headers.set('experimental', 'true')
@@ -189,8 +188,8 @@ export class CasesService {
     return this.http
       .get(url, {headers, observe: 'body'})
       .pipe(
-        map(response => {
-          return this.handleNestedDynamicLists(response);
+        map(body => {
+          return this.handleNestedDynamicLists(body);
         }),
         catchError(error => {
           this.errorService.setError(error);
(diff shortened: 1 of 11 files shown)


src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -103,19 +102,19 @@ export class CasesService {
    *
    * EUI-2530 Dynamic Lists for Elements in a Complex Type
    *
-   * @param jsonResponse - {}
+   * @param jsonBody - {}
    */
-  private handleNestedDynamicLists(jsonResponse) {
+  private handleNestedDynamicLists(jsonBody: any): any {
```
Daniel-Lam, line 107: @paul-graham given that `jsonBody` is listed as `{}` in the params, would it work if you declared the same in the method parameter, i.e. `jsonBody: {}`, as opposed to just `any`?

src/shared/services/workbasket/workbasket-input-filter.service.ts
```diff
@@ -32,9 +32,8 @@ export class WorkbasketInputFilterService {
     this.currentCaseType = caseTypeId;
     return this.httpService
       .get(url, {headers, observe: 'body'})
-      .map(response => {
-        let jsonResponse = response;
-        let workbasketInputs = jsonResponse.workbasketInputs;
+      .map(body => {
+        let workbasketInputs = body.workbasketInputs;
```
Daniel-Lam, line 36: My favourite thing... can this be a `const` instead of a `let`? 😄

Daniel-Lam (changes requested):
@paul-graham - just two review comments to consider; otherwise, all looks OK

src/shared/services/workbasket/workbasket-input-filter.service.ts
paul-graham: Done. Pushed.

src/shared/components/case-editor/services/cases.service.ts
paul-graham: That won't work but I'll do one better as you're right - `any` is kinda pants.

paul-graham: New version pushed with a better type than `any`. 🤦

Daniel-Lam (approved):

src/shared/services/banners/banners.service.ts
```diff
@@ -14,19 +14,15 @@ export class BannersService {
...
 
   getBanners(jurisdictionReferences: string[]): Observable<Banner[]> {
-    let url = this.appConfig.getBannersUrl();
-    let headers = new HttpHeaders()
+    const url = this.appConfig.getBannersUrl();
+    const headers = new HttpHeaders()
       .set('experimental', 'true')
       .set('Accept', BannersService.V2_MEDIATYPE_BANNERS)
       .set('Content-Type', 'application/json');
     let params = new HttpParams();
```
kuda1992, line 22: params could be const as well.

src/shared/services/document-management/document-management.service.ts
```diff
@@ -29,9 +29,7 @@ export class DocumentManagementService {
       .post(url, formData, {headers, observe: 'body'})
       .pipe(delay(DocumentManagementService.RESPONSE_DELAY))
       .pipe(
-        map(response => {
-          return response;
-        })
+        map(body => body) // EUI-3495. Not sure we need this?
```
kuda1992, line 32: This is suspect? Do we need a map here?

kuda1992 (approved):
LGTM

src/shared/services/banners/banners.service.ts
paul-graham: That was my first reading of it too. However, the `forEach()` in the line below reassigns it each time so it needs to remain `let`. I'm not 100% sure why the reassignment is needed but it was done deliberately by @udaydenduluri33 in the first place - and is the same in some other places that `HttpHeaders` is used - so I assume it's necessary. 🤷

src/shared/services/document-management/document-management.service.ts
paul-graham: Hence my comment question.

It's inside the `pipe()`, though, and @Daniel-Lam recently found some peculiarities around the document upload functionality so I didn't want to make a change to it at this point as it _is_ working the way it is. I simply made the code clearer for the time being by replacing `response` with `body` and removing the unnecessary code block.

For the moment, I'd prefer to leave it as it is and have another look at it at some point in the future - it _looks_ like a pointless operation, though, I agree.

src/shared/services/banners/banners.service.ts
kuda1992: The reassignment is not necessary at all. But maybe @udaydenduluri33 has an explanation?

src/shared/services/document-management/document-management.service.ts
kuda1992: Mapping `body => body` does nothing. It's fair if there is a reason.

src/shared/services/banners/banners.service.ts
udaydenduluri33: I remember helping @RiteshHMCTS  on this for his ticket. This is because HttpHeaders are immutable.

paul-graham: Ah! That would explain it then. I hadn't appreciated that and I've just checked the Angular docs to confirm there's no alternative mechanism... but @udaydenduluri33's explanation looks correct from what I can see. Happy to resolve this conversation, @kuda1992?

kuda1992: @udaydenduluri2 @paul-graham just checked out the documentation. its apparently correct. I will resolve the conversation.

src/shared/services/document-management/document-management.service.ts
paul-graham: Done and pushed.

udaydenduluri33 (approved):
LGTMp

PR #708, opened 2021-02-25
EUI-3468-Text-field-disabled

ishOsw:
- [x] README and other documentation has been updated / added (if needed)

https://tools.hmcts.net/jira/browse/EUI-3468

### Change description ###
fix for disabled textfield

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -17,6 +17,8 @@ import { DRAFT_PREFIX } from '../../../domain/draft.model';
 import { Wizard } from '../domain/wizard.model';
 import { CaseField } from '../../../domain/definition';
 import { FieldsUtils } from '../../../services/fields';
+import { DataRowOutlet } from '@angular/cdk/table';
+import { tsStructureIsReused } from '@angular/compiler-cli/src/transformers/util';
 
 @Component({
   selector: 'ccd-case-edit-page',
@@ -91,6 +93,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
 
   applyValuesChanged(valuesChanged: boolean): void {
     this.formValuesChanged = valuesChanged;
+    this.currentPageIsNotValid();
   }
 
   private initDialog() {
@@ -113,7 +116,33 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
   }
 
   currentPageIsNotValid(): boolean {
-    return !this.pageValidationService.isPageValid(this.currentPage, this.editForm);
+    const data = this.editForm.controls.data['controls'];
+    if (data) {
+      if (this.currentPage.id === 'editAppealnewMatters') {
+        if (data.newMatters && data.hasNewMatters) {
+          const newMatters = data.newMatters.value ? data.newMatters.value : '';
+          const hasNewMatters = data.hasNewMatters.value ? data.hasNewMatters.value : '';
+          if (newMatters.length === 0 && hasNewMatters === 'Yes') {
+            return true;
+          }
+        }
+      } else {
+        if (data.email && data.mobileNumber) {
+          const email = data.email.value ? data.email.value : '';
+          const mobileNumber = data.mobileNumber.value ? data.mobileNumber.value : '';
+          if (email.length === 0 && mobileNumber.length === 0) {
+            return true;
+          }
+          if (email.length !== 0 && mobileNumber.length === 0) {
+            return false;
+          }
+          if (email.length === 0 && mobileNumber.length !== 0) {
+            return false;
+          }
+        }
+      }
+    }
+   return !this.pageValidationService.isPageValid(this.currentPage, this.editForm);
   }
 
   toPreviousPage() {
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
@@ -17,8 +17,6 @@ import { DRAFT_PREFIX } from '../../../domain/draft.model';
 import { Wizard } from '../domain/wizard.model';
 import { CaseField } from '../../../domain/definition';
 import { FieldsUtils } from '../../../services/fields';
-import { DataRowOutlet } from '@angular/cdk/table';
-import { tsStructureIsReused } from '@angular/compiler-cli/src/transformers/util';
 
 @Component({
   selector: 'ccd-case-edit-page',
--- a/src/shared/components/palette/email/write-email-field.component.ts
+++ b/src/shared/components/palette/email/write-email-field.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { FormControl } from '@angular/forms';
+import { FormControl} from '@angular/forms';
 import { AbstractFieldWriteComponent } from '../base-field/abstract-field-write.component';
 
 @Component({
--- a/src/shared/components/palette/email/write-email-field.component.ts
+++ b/src/shared/components/palette/email/write-email-field.component.ts
@@ -11,7 +11,7 @@ export class WriteEmailFieldComponent extends AbstractFieldWriteComponent implem
   emailControl: FormControl;
 
   ngOnInit() {
-    let notEmpty = this.caseField.value !== null && this.caseField.value !== null;
+    let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
     this.emailControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null)) as FormControl;
     if (this.emailControl.disabled) {
       this.emailControl.enable({emitEvent: false});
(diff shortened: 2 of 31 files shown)


leepavlouhmcts (approved):
looks good

rayliangatos:
I think above changes probably do not fulfil what is required from the reporter, could you contact with @rohith tiru or demo what you have done to him to see whether he agreed with your fix.

In the description of the ticket what I understand is user don't want the 'Continue' button to be enabled if user do not enter valid answer. It's not related to whether these text/email is enabled/disabled.

src/shared/components/palette/email/write-email-field.component.ts
```diff
@@ -11,6 +11,10 @@ export class WriteEmailFieldComponent extends AbstractFieldWriteComponent implem
   emailControl: FormControl;
 
   ngOnInit() {
-    this.emailControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
```
Daniel-Lam, line 14: Should be a `const` rather than `let`

src/shared/components/palette/text-area/write-text-area-field.component.ts
```diff
@@ -16,7 +16,11 @@ export class WriteTextAreaFieldComponent extends AbstractFieldWriteComponent imp
   }
 
   ngOnInit() {
-    this.textareaControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
```
Daniel-Lam, line 19: Should be a `const` rather than `let`

src/shared/components/palette/text/write-text-field.component.ts
```diff
@@ -11,6 +11,10 @@ export class WriteTextFieldComponent extends AbstractFieldWriteComponent impleme
   textControl: FormControl;
 
   ngOnInit() {
-    this.textControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
```
Daniel-Lam, line 14: Should be a `const` rather than `let`

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -17,6 +17,8 @@ import { DRAFT_PREFIX } from '../../../domain/draft.model';
 import { Wizard } from '../domain/wizard.model';
 import { CaseField } from '../../../domain/definition';
 import { FieldsUtils } from '../../../services/fields';
+import { DataRowOutlet } from '@angular/cdk/table';
```
Daniel-Lam, line 20: Is this import being used? Remove if not

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -17,6 +17,8 @@ import { DRAFT_PREFIX } from '../../../domain/draft.model';
...
+import { tsStructureIsReused } from '@angular/compiler-cli/src/transformers/util';
```
Daniel-Lam, line 21: Same for this import as above

Daniel-Lam (changes requested):
Hi @ishOsw - a few changes to be made. Also, you need to add unit tests for the `currentPageIsNotValid()` function in `case-edit-page.component.ts` and, if possible, update or extend any applicable unit tests for:
`write-email-field.component.ts`
`write-text-area-field.component.ts`
`write-text-field.component.ts`

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -91,6 +91,7 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
 
   applyValuesChanged(valuesChanged: boolean): void {
     this.formValuesChanged = valuesChanged;
+    this.currentPageIsNotValid();
```
paul-graham, line 94: Not sure why this is being called here as it doesn't capture the returned value or do anything with it...?

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -113,7 +114,33 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
   }
 
   currentPageIsNotValid(): boolean {
-    return !this.pageValidationService.isPageValid(this.currentPage, this.editForm);
+    const data = this.editForm.controls.data['controls'];
+    if (data) {
+      if (this.currentPage.id === 'editAppealnewMatters') {
```
paul-graham, line 119: This is _super_ specific and means you're only fixing it for a single event trigger. If there's another event with the same structure - or they rename the event - this code won't be called. I don't think the toolkit should have handling for specific events like this.

paul-graham: All of the calls to `newMatters` and `hasNewMatters` below are also specific to this very one trigger event.

src/shared/components/case-editor/case-edit-page/case-edit-page.component.ts
```diff
@@ -113,7 +114,33 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
...
+        if (data.newMatters && data.hasNewMatters) {
+          const newMatters = data.newMatters.value ? data.newMatters.value : '';
+          const hasNewMatters = data.hasNewMatters.value ? data.hasNewMatters.value : '';
+          if (newMatters.length === 0 && hasNewMatters === 'Yes') {
+            return true;
+          }
+        }
+      } else {
+        if (data.email && data.mobileNumber) {
```
paul-graham, line 128: This whole block now takes no account of how the CCD Case config is actually set up - I've seen configurations where there was an optional `email` and an optional `mobileNumber` field. This code will make that look invalid when it shouldn't.

src/shared/components/palette/email/write-email-field.component.ts
```diff
@@ -11,6 +11,10 @@ export class WriteEmailFieldComponent extends AbstractFieldWriteComponent implem
   emailControl: FormControl;
 
   ngOnInit() {
-    this.emailControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    const notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
+    this.emailControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null)) as FormControl;
+    if (this.emailControl.disabled) {
```
paul-graham, line 16: This takes no account of why the control has been disabled. The conditional show/hide logic will hide and disable fields that shouldn't be shown so that the data from those hidden fields doesn't get submitted. If you've added code to simply enable anything that's disabled - which is what this does - there's every chance the submitted data will be invalid or, at least, that you'll be submitting data for fields that are hidden and shouldn't be submitted.

src/shared/components/palette/text-area/write-text-area-field.component.ts
```diff
@@ -16,7 +16,11 @@ export class WriteTextAreaFieldComponent extends AbstractFieldWriteComponent imp
   }
 
   ngOnInit() {
-    this.textareaControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    const notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
+    this.textareaControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null)) as FormControl;
+    if (this.textareaControl.disabled) {
```
paul-graham, line 21: Same as above:
This takes no account of why the control has been disabled. The conditional show/hide logic will hide and disable fields that shouldn't be shown so that the data from those hidden fields doesn't get submitted. If you've added code to simply enable anything that's disabled - which is what this does - there's every chance the submitted data will be invalid or, at least, that you'll be submitting data for fields that are hidden and shouldn't be submitted.

src/shared/components/palette/text/write-text-field.component.ts
```diff
@@ -11,6 +11,10 @@ export class WriteTextFieldComponent extends AbstractFieldWriteComponent impleme
   textControl: FormControl;
 
   ngOnInit() {
-    this.textControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    const notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
+    this.textControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null)) as FormControl;
+    if (this.textControl.disabled) {
```
paul-graham, line 16: And again:
This takes no account of why the control has been disabled. The conditional show/hide logic will hide and disable fields that shouldn't be shown so that the data from those hidden fields doesn't get submitted. If you've added code to simply enable anything that's disabled - which is what this does - there's every chance the submitted data will be invalid or, at least, that you'll be submitting data for fields that are hidden and shouldn't be submitted.

paul-graham (changes requested):
The `currentPageIsNotValid()` method cannot get merged into `master` the way it's currently written as it is simultaneously incredibly specific to a single event and then ignores the case configuration for anything _other_ than that single event.

The CCD Case UI Toolkit needs to consider the case configuration carefully and act accordingly on that basis. The code in that method doesn't consider the configuration at all and it would be almost _guaranteed_ to result in a bug about a regression being raised within days of its release.

I haven't looked at the configuration of the case in question but _either_:
1. There's something in the case configuration you can use to handle this situation in a generic manner, or;
2. The case configuration itself is broken and there's nothing that needs to be done within the toolkit to handle it - the responsible service just needs to change their case configuration.

PR #712, opened 2021-03-03
Bug/e UI 3466 date fallback

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-3466

### Change description ###

Implemented fallback for date missing 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-list/case-list.component.ts
+++ b/src/shared/components/case-list/case-list.component.ts
@@ -43,11 +43,11 @@ export class CaseListComponent {
   constructor(private browserService: BrowserService) { }
 
   public formatDate(date: Date): string {
-    return formatDate(date, 'dd MMM yyyy', 'en-GB');
+    return date ? formatDate(date, 'dd MMM yyyy', 'en-GB') : '-';
   }
 
   public formatDateAtTime(date: Date): string {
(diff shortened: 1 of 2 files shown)


Daniel-Lam (approved):
Changes look OK, @adnanakgun - approved

src/shared/components/case-list/case-list.component.ts
```diff
@@ -43,11 +43,11 @@ export class CaseListComponent {
   constructor(private browserService: BrowserService) { }
 
   public formatDate(date: Date): string {
-    return formatDate(date, 'dd MMM yyyy', 'en-GB');
+    return date ? formatDate(date, 'dd MMM yyyy', 'en-GB') : '-';
```
paul-graham, line 46: Why "-"? I didn't see that mentioned on the ticket... 🤔

paul-graham (changes requested):

src/shared/components/case-list/case-list.component.ts
adnanakgun: The ticket describes a bug which is fixed by the introduction of a fallback for date fields.

paul-graham (approved):

PR #718, opened 2021-03-08
EUI-3267: Reintroducing grey bars

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-3267 ✅ `DONE`

### Change description ###
Reintroduced grey bars for fields that are conditionally shown, depending on the value of other fields. There are some exceptions in place for collections, and fields that depend solely on hidden or readonly fields (showing a grey bar for them looks odd and is also in line with what appeared to be in place previously).

In line with what was previously in place - based on what I could find - the functionality doesn't chain - i.e., if Field A is set to "Yes" and that displays Field B, Field B will have a grey bar. If Field B is then set to "Yes" and that displays Field C, Field C will also show a grey bar, indented to the same level is Field B, rather than even further indented or showing multiple grey bars.

```
[ ] Yes
[x] No
```


src/shared/components/palette/base-field/field-write.component.ts
```diff
@@ -58,5 +58,9 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements
     }
     component.instance['isExpanded'] = this.isExpanded;
     this.fieldContainer.insert(component.hostView);
+
+    // EUI-3267.
+    // Set up the flag for whether this can have a grey bar.
+    this.canHaveGreyBar = this.caseField.show_condition && this.caseField.field_type.type !== 'Collection';
```
udaydenduluri33, line 64: Do we need to check for truthy of caseField? Or its always populated?

src/shared/directives/conditional-show/domain/conditional-show.model.ts
```diff
@@ -67,6 +68,54 @@ export class ShowCondition {
...
   }
+
+  /**
+   * Determine whether this is affected by fields that have a display_context
+   * of HIDDEN or READONLY, which means they aren't able to be changed by the
+   * user's actions.
+   *
+   * @param caseFields Inspected to see appropriate display_context(s).
+   */
+  public hiddenCannotChange(caseFields: CaseField[]): boolean {
```
udaydenduluri33, line 79: I understand lots of logic in this method hiddenCannotChange.Can we have as unit test to it?

src/shared/components/palette/base-field/field-write.component.ts
paul-graham: Theoretically, but if `caseField` is null, it will have gone wrong before it gets to this bit of code as prior lines use properties of it without a null check. There's also code above to instantiate a `CaseField` if it's not of the correct type.

src/shared/directives/conditional-show/domain/conditional-show.model.ts
paul-graham: Absolutely! Just added 23 unit tests to cover all the scenarios I could think of - happy and unhappy paths.

udaydenduluri33 (approved):
LGTMp

Mo-Lala84 (approved):
approved

PR #719, opened 2021-03-08
Release/2.74.0

adnanakgun:
Release for issues

EUI-2778- Dynamic list incorrect format for mid event callbacks
EUI-2830- Fix dynamic lists in collections
EUI-2985- Facing an issue with CCD UI where some of the fields are getting shown as disabled

### DIFF
--- a/src/shared/components/palette/fixed-radio-list/write-fixed-radio-list-field.component.ts
+++ b/src/shared/components/palette/fixed-radio-list/write-fixed-radio-list-field.component.ts
@@ -1,17 +1,21 @@
 import { Component, OnInit } from '@angular/core';
 import { FormControl } from '@angular/forms';
 import { AbstractFieldWriteComponent } from '../base-field/abstract-field-write.component';
-
+​
 @Component({
   selector: 'ccd-write-fixed-radio-list-field',
   templateUrl: './write-fixed-radio-list-field.html'
 })
 export class WriteFixedRadioListFieldComponent extends AbstractFieldWriteComponent implements OnInit {
-
+​
   fixedRadioListControl: FormControl;
-
+​
   ngOnInit() {
     let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
     this.fixedRadioListControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null)) as FormControl;
+    if (this.fixedRadioListControl.disabled) {
+      this.fixedRadioListControl.enable({emitEvent: false});
+    }
   }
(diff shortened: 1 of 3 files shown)


Daniel-Lam (changes requested):
@adnanakgun - all looks OK but do you need to update `RELEASE-NOTES.md`?

src/shared/components/palette/fixed-radio-list/write-fixed-radio-list-field.component.ts
```diff
@@ -1,17 +1,21 @@
...
-
+​
   fixedRadioListControl: FormControl;
-
+​
   ngOnInit() {
     let notEmpty = this.caseField.value !== null && this.caseField.value !== undefined;
     this.fixedRadioListControl = this.registerControl(new FormControl(notEmpty ? this.caseField.value : null)) as FormControl;
+    if (this.fixedRadioListControl.disabled) {
+      this.fixedRadioListControl.enable({emitEvent: false});
```
paul-graham, line 17: What is this bit for? There is logic to disable fields when they get hidden. Theoretically, _this_ should happen first... but I'm not sure why it's there at all...?

paul-graham (changes requested):

src/shared/components/palette/fixed-radio-list/write-fixed-radio-list-field.component.ts
adnanakgun: I simply copied from the solution for EUI-2985 developed by @ishOsw.

paul-graham: Argh! I didn't see a PR for EUI-2985 or I'd have raised this same thing there! The code looks risky to me because it takes no account of _why_ the control was disabled in the first place - it just blanket re-enables a disabled control.

Has it been tested with hidden fields, @ishOsw? Did you investigate the cause of the disabling to determine whether it's safe to re-enable it?

At the very least, there should be a check for the `hidden` property in the `CaseField` to make sure re-enabling the control doesn't contradict that - i.e., add `&& this.caseField.hidden !== true` to the `if` statement.

paul-graham: Update: the original change was from a while ago. It needs to take account of `this.caseField.hidden` now, though, and only re-enable if it's not `true`. That said, something _else_ might be legitimately disabling the control. 🤷‍♂️

PR #723, opened 2021-03-10
profile call needs to happen before event trigger

RiteshHMCTS:
https://tools.hmcts.net/jira/browse/EUI-3335

### Change description ###
Make profile call before event trigger 

```
[ ] Yes
[x ] No
```


src/shared/components/case-viewer/services/event-trigger.resolver.ts
```diff
@@ -40,6 +42,7 @@ export class EventTriggerResolver implements Resolve<CaseEventTrigger> {
     if (-1 === EventTriggerResolver.IGNORE_WARNING_VALUES.indexOf(ignoreWarning)) {
       ignoreWarning = 'false';
     }
+    this.profileService.get().subscribe(_ => this.profileNotifier.announceProfile(_));
```
phillip-whitaker, line 45: If this profile call truly needs to be done before the case loading, I think it has to be pipe'd in. Maybe using a forkJoin with the case service call.

RiteshHMCTS: I have replaced the http calls with forkJoin

phillip-whitaker (dismissed):

adnanakgun (dismissed):

src/shared/components/case-viewer/services/event-trigger.resolver.ts
```diff
@@ -40,10 +42,17 @@ export class EventTriggerResolver implements Resolve<CaseEventTrigger> {
     if (-1 === EventTriggerResolver.IGNORE_WARNING_VALUES.indexOf(ignoreWarning)) {
       ignoreWarning = 'false';
     }
-    return this.casesService
-      .getEventTrigger(caseTypeId, eventTriggerId, cid, ignoreWarning)
+
+    const profileObserver = this.profileService.get();
+    const eventTriggerObserver = this.casesService.getEventTrigger(caseTypeId, eventTriggerId, cid, ignoreWarning);
+
+    return forkJoin([profileObserver, eventTriggerObserver])
```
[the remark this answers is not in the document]
andywilkinshmcts: I think you need to put the call to getEventTrigger in the handler for the profileService, something like this (very rough, not compiled or tested but you get the idea)
profileObserver
.pipe(
map(profileDate) => {
  this.profileNotifier.announceProfile(profileData);
  return this.casesService
      .getEventTrigger(caseTypeId, eventTriggerId, cid, ignoreWarning)
      .pipe (
      map (eventTrigger => this.cachedEventTrigger = eventTrigger))

andywilkinshmcts (changes requested):

src/shared/components/case-viewer/services/event-trigger.resolver.ts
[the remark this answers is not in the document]
RiteshHMCTS: made the first call to profiler and then to case service.

phillip-whitaker (approved):

kuda1992 (dismissed):
looks good

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "3.1.5-complex-field-check-answers",
+  "version": "3.1.5-profile-call-before-event-trigger",
```
andywilkinshmcts, line 3: Is this correct? On main the current version  "3.1.5-complex-field-check-answers"
Should we not be incrementing the version

andywilkinshmcts (changes requested):
Check with mo what version number to use

kuda1992 (approved):

package.json
RiteshHMCTS: we don't do version increment only postfix ticket reference and the same confirmed with @Mo-Lala84

andywilkinshmcts (approved):

PR #724, opened 2021-03-10
EUI-2667: DatePipe fix

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2667 ✅ `DONE`

### Change description ###
The `DatePipe` was throwing exceptions for dates that didn't fit the ISO format. This resulted in console errors and a failure to display dates when they are in a collection - the date is basically formatted twice, except the second time it throws an exception because it receives the already-formatted date.

This will now attempt to parse dates it receives that don't correspond to the ISO format and, when it encounters one, format it appropriately.

```
[ ] Yes
[x] No
```

**Notes**
This is a Draft PR for the moment as the original went stale after not being merged, despite the JIRA having been tested and approved. I've now rebased it against `master` so it should be re-tested.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,8 @@
 ## RELEASE NOTES
 
+### Version 2.73.2-collection-date-field
+**EUI-2667** Fix to date field rendering within collections
+
 ### Version 2.73.0
 **EUI-3222** Loading spinner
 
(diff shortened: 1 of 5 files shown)


andywilkinshmcts (approved):

paulhowes-HMCTS (approved):
Looks ok, and seeing as it has already been approved prior to going stale then all good!

PR #725, opened 2021-03-10
Non-descriptive heading amends for accessibility (AA)

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-3506

### Change description ###
Non-descriptive heading amends for accessibility (AA)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/case-edit-page/case-edit-page.html
+++ b/src/shared/components/case-editor/case-edit-page/case-edit-page.html
@@ -1,5 +1,10 @@
-<!-- Current Page && Event trigger name -->
-<h1 class="heading-h1"><span *ngIf="currentPage && currentPage.label">{{currentPage.label}}&nbsp;-&nbsp;</span>{{ eventTrigger.name}}</h1>
+<ng-container *ngIf="currentPage">
+  <h1 *ngIf="!currentPage.label" class="govuk-heading-l">{{eventTrigger.name}}</h1>
+  <ng-container *ngIf="currentPage.label">
+    <span class="govuk-caption-l">{{ eventTrigger.name}}</span>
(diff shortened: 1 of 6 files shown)


src/shared/components/case-editor/case-edit-page/case-edit-page.html
```diff
@@ -1,5 +1,10 @@
-<!-- Current Page && Event trigger name -->
-<h1 class="heading-h1"><span *ngIf="currentPage && currentPage.label">{{currentPage.label}}&nbsp;-&nbsp;</span>{{ eventTrigger.name}}</h1>
+<ng-container *ngIf="currentPage">
+  <h1 *ngIf="!currentPage.label" class="govuk-heading-l">{{eventTrigger.name}}</h1>
+  <ng-container *ngIf="currentPage.label">
+    <span class="govuk-caption-l">{{ eventTrigger.name}}</span>
```
paul-graham, line 4: These look to be in a different order to the original of `currentPage.label` followed by `eventTrigger.name` (and also separated by a hyphen). Is the change in order and lack of a hyphen intentional?

paul-graham (changes requested):
Just the one thing to highlight...

src/shared/components/case-editor/case-edit-page/case-edit-page.html
paulhowes-HMCTS: Yes - it's basically being split in two. If there is no label then the eventTrigger is the heading, and if there is a label the label becomes the heading and the eventTrigger becomes the gov-uk caption that sits above the heading.

paul-graham (approved):
Given the response to my question, I'm happy that everything here looks fine. 👍

phillip-whitaker (approved):

PR #730, opened 2021-03-15
EUI-3594: Error message specifics not being displayed

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-3594

### Change description ###
Fix issue with HTTP error handling post-Angular `HttpClient` upgrade, which failed to map error properties between new `HttpErrorResponse` objects being returned and `HttpError` domain model objects for UI display.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-viewer/case-viewer.component.spec.ts
+++ b/src/shared/components/case-viewer/case-viewer.component.spec.ts
@@ -1,3 +1,4 @@
+import { HttpErrorResponse } from '@angular/common/http';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { Component, DebugElement, EventEmitter, Input, Output } from '@angular/core';
 import { CaseViewerComponent } from './case-viewer.component';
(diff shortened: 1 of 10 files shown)


src/shared/components/case-viewer/case-viewer.component.spec.ts
```diff
@@ -1,3 +1,4 @@
+import { HttpErrorResponse } from '@angular/common/http';
```
paul-graham, line 1: I'm seeing an `xdescribe(...)` for all the tests in this suite...?

paul-graham (changes requested):
Just that one question in `case-viewer.component.spec.ts` where it looks like there are changes but all the tests are actually being skipped. Not sure if you were aware of it, @Daniel-Lam, so this is to flag it up for your attention rather than _actually_ request changes.

src/shared/components/case-viewer/case-viewer.component.spec.ts
Daniel-Lam: @paul-graham - that's a bit odd... I tried an `fdescribe` and all those tests fail! They ought to be fixed but that would be outside the scope of this PR, so I've left as `xdescribe` (as it was) for now.

paul-graham: Okay, cool. So long as you're aware of it. We should create a tech debt ticket to tidy up any tests that have an `x` prefix (`xdescribe` or `xit`) - either fix them or remove them. 😄

paul-graham (approved):
The `xdescribe` hasn't been introduced for this PR and @Daniel-Lam was aware it was there so I'm happy to approve these changes.

adnanakgun (approved):

PR #731, opened 2021-03-17
EUI-3146-numeric-comparator-hot-fix

rayliangatos:

https://tools.hmcts.net/jira/browse/EUI-3146
https://tools.hmcts.net/jira/browse/RDM-11029

### Change description ###
An edge case when the CCD configs return Case ID for sorting, it's a numeric field not string tyoe, the sort itself at least need to be compatible for string/numeric sorting, right now it only consider string so it failed on .toLowerCase

```
[ ] Yes
[x] No
```

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,8 @@
 ## RELEASE NOTES
 
+### Version 2.73.6-numeric-comparator
+**EUI-3146** numeric-comparator-hot-fix
+
 ### Version 2.73.5-ngmodel-deprecation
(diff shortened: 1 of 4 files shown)


paul-graham (changes requested):
Looks like exactly the same issue as reported in https://tools.hmcts.net/jira/browse/EUI-3608.

What happens when you click on the column header to sort by that column? When I made a very similar change in the midst of investigating EUI-3608, CCD threw an exception. You don't have any code in here to alter the call being made to CCD so you'll almost certainly get precisely the same result.

The issue is with the configuration of the columns, not with this code.

PR #733, opened 2021-03-19
Add support for real money types. 

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-3583

### Change description ###
`ReadMoneyGbpFieldComponent 
` should support real values. Which means both negative and positive values should be treated as valid form fields.

```
[ ] Yes
[ x] No
```


src/shared/components/palette/base-field/abstract-form-field.component.ts
```diff
@@ -16,7 +16,7 @@ export abstract class AbstractFormFieldComponent {
   @Input()
   parent?: FormContainer;
 
-  protected registerControl<T extends AbstractControl>(control: T, replace = false): AbstractControl {
+  public registerControl<T extends AbstractControl>(control: T, replace = false): AbstractControl {
```
paul-graham, line 19: Why has this been made public? If it's for testing purposes, that necessity had already been removed.

src/shared/components/palette/money-gbp/read-money-gbp-field.component.spec.ts
```diff
@@ -175,19 +226,19 @@ describe('ReadMoneyGBPFieldComponent', () => {
         .compileComponents();
 
       fixture = TestBed.createComponent(ReadMoneyGbpFieldComponent);
-      component = fixture.componentInstance;
+      component = fixture.componentInstance as ReadMoneyGbpFieldComponent;
 
+      component.registerControl = REGISTER_CONTROL;
       component.amount = undefined;
       component.caseField = CASE_FIELD;
-      component.formGroup = FORM_GROUP;
```
paul-graham, line 182: Why have you changed this mechanism? It works everywhere else and the `registerControl` method shouldn't be public as it's only ever invoked from within the component, not externally.

src/shared/components/palette/money-gbp/write-money-gbp-field.component.spec.ts
```diff
@@ -52,8 +125,8 @@ describe('WriteMoneyGbpFieldComponent', () => {
     fixture = TestBed.createComponent(WriteMoneyGbpFieldComponent);
     component = fixture.componentInstance;
 
+    component.registerControl = REGISTER_CONTROL;
     component.caseField = CASE_FIELD;
-    component.formGroup = FORM_GROUP;
```
paul-graham, line 56: Same question as above. This works elsewhere so why has it been replaced with an old mechanism?

paul-graham (changes requested):

src/shared/components/palette/base-field/abstract-form-field.component.ts
kuda1992: Well I made it public to test the method. What do you suggest keep it protected?

src/shared/components/palette/money-gbp/read-money-gbp-field.component.spec.ts
kuda1992: Honest truth I have changed the mechanism. I just replicated the changes made from the previous PR.

src/shared/components/palette/base-field/abstract-form-field.component.ts
paul-graham: Yes please. It was made protected because it's only ever called from within the component (and anything that extends this base component) and never externally.

src/shared/components/palette/money-gbp/write-money-gbp-field.component.spec.ts
kuda1992: Same answer as before. Just copied the changes from the PR.

src/shared/components/palette/money-gbp/read-money-gbp-field.component.spec.ts
paul-graham: Ah! That would explain it! Yes, that got replaced with the mechanism of setting `component.formGroup`, which did away with the need for a mock `registerControl` method and also the need for the _actual_ `registerControl` method being public. Can you undo these bits of changes, please? :)

paul-graham (dismissed):
Good stuff. Thank you for making the requested changes, @kuda1992!

Approved. 👍

udaydenduluri33 (dismissed):
LGTMk

PR #738, opened 2021-03-26
Replace h2 tags with spans on case edit pages (A)

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-3550

### Change description ###
Replace h2 tags with spans on case edit pages for accessibility

```
[ ] Yes
[x] No
```


src/shared/components/case-editor/case-edit-submit/case-edit-submit.html
```diff
@@ -2,9 +2,9 @@
 <h1 class="govuk-heading-l">{{ eventTrigger.name}}</h1>
 
 <!-- Case ID -->
-<h2 *ngIf="getCaseId()" class="heading-h2">
+<span *ngIf="getCaseId()" class="heading-span" title="Case reference {{ getCaseId() | ccdCaseReference}}">
   #{{ getCaseId() | ccdCaseReference}}
-</h2>
+</span>
```
phillip-whitaker, line 7: I really don't like this <span> thing... I know it is what you have been told to do. But it is such an abuse of spans

phillip-whitaker (approved):

src/shared/components/case-editor/case-edit-submit/case-edit-submit.html
phillip-whitaker: I do think this will get flagged in future accessibility audits for poor HTML structure

paulhowes-HMCTS: It highlights the need to move the case reference somewhere else, however for the moment 'demoting' the h2 to a span is a step in the right direction.

paul-graham (dismissed):
Agree with @phillip-whitaker about the abuse of `<span>`s. And it strikes me as peculiar that it's been decided that the title should include the case reference as it'll mean screen readers announce the case reference twice in succession, separated merely by the word "number" for the "#" within the `<span>`. 🤷‍♂️

Given you've pretty much overridden the styling of `.heading-h2` from MC, I don't _think_ this needs to be considered a breaking change. Opinions may vary on this one, though.

paulhowes-HMCTS:
Replaced by https://github.com/hmcts/ccd-case-ui-toolkit/pull/783 as part of CT_v1.0.1 - Release 22.1 - Candidate

PR #739, opened 2021-03-27
EUI-3587: Fix "Print URL" pipe for rewriting remote Print Service URLs

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-3587

### Change description ###
Fix the pipe to rewrite remote Print Service URLs to the correct "local" version to be used for the front-end, irrespective of environment-based variations of the remote domain name.

```
[ ] Yes
[x] No
```

### DIFF
--- a/demo/src/app/app.config.ts
+++ b/demo/src/app/app.config.ts
@@ -22,7 +22,6 @@ export class AppConfig extends AbstractAppConfig {
     'activity_url': '',
     'activity_max_request_per_batch': 25,
     'print_service_url': '/print',
-    'remote_print_service_url': '/remote_print',
     'pagination_page_size': 25,
     'prd_url': 'api/caseshare/orgs',
     'cache_time_out': 45000,
@@ -72,7 +71,7 @@ export class AppConfig extends AbstractAppConfig {
   public getPaymentsUrl() {
     return this.config.payments_url;
   }
-  
+
   public getPayBulkScanBaseUrl() {
     return this.config.pay_bulk_scan_url;
   }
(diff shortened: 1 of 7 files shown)


demo/src/app/app.config.ts
```diff
@@ -116,10 +115,6 @@ export class AppConfig extends AbstractAppConfig {
     return this.config.print_service_url;
   }
 
-  public getRemotePrintServiceUrl() {
```
paul-graham, line 119: I think this makes this a breaking change and so it would need to go to the next major release...

paul-graham (changes requested):
I think this needs to be a major release: `3.0.0`.

adnanakgun (approved):
I agree that this is a breaking change. Approving subject to @paul-graham 's change request.

demo/src/app/app.config.ts
Daniel-Lam: I've now re-introduced this in `app.config.ts` as a deprecated, no-op function that returns `undefined`, ahead of its removal in the next major release.

src/app.config.ts
```diff
@@ -47,7 +54,8 @@ export class CaseEditorConfig {
   activity_url: string;
   activity_max_request_per_batch: number;
   print_service_url: string;
-  remote_print_service_url: string;
+  // remote_print_service_url marked as optional since deprecation, ahead of removal in next major release
```
paul-graham, line 57:
```
/**
 * @deprecated?
 */
```

Daniel-Lam: Good point - done

paul-graham (approved):
Thank you for making the changes necessary to stop this being a breaking change, @Daniel-Lam. As you've pointed out, version `3.0.0` is earmarked for much more exciting things and this would have put an unnecessary obstacle in the way of that.

Given the deprecation flags and the other changes made, happy to approve this. 👍

paul-graham (approved):
Changes to the `package.json` and `RELEASE-NOTES.md` look all good. Re-approving. 👍

adnanakgun (approved):

PR #769, opened 2021-05-05
Feature/e UI 3049 datetime picker final

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-3049
https://tools.hmcts.net/jira/browse/EUI-3051

### Change description ###
New WriteDateContainerComponent components displays DatetimePickerComponent or WriteDateFieldComponent based on the dateTimeEntryFormat field

```
[ ] Yes
[ x ] No
```


udaydenduluri33 (dismissed):
LGTMc

src/shared/components/palette/utils/date.pipe.spec.ts
```diff
@@ -51,12 +51,14 @@ describe('DatePipe', () => {
     expect(message).toBe('26 Jul 2017, ' + getExpectedHour(12) + ':09:05 PM');
   });
 
-  it('should render correct date if UTC date in yyyy-mm-ddThh:mm:ss PM format', () => {
+  /* requirement to no longer have effectively changing timezones negates test
```
paul-graham, line 54: _Question:_
I'm not sure if this comment implies that the test is dead and should be removed (if so, it should be removed rather than commented out) or if it's suggesting that there's a change in the requirements that means this test needs to change too...?

src/shared/components/palette/utils/date.pipe.spec.ts
```diff
@@ -157,18 +159,7 @@ describe('DatePipe', () => {
...
-  it ('should handle GMT to BST transition', () => {
-    let endOfWinter = new Date(2020, 2, 29, 0, 59, 59)
-    let message = datePipe.transform (endOfWinter.toISOString(), 'GMT', 'dd MMMM yyyy HH:mm:ss.SSS');
-    expect(message).toBe('29 March 2020 00:59:59.000');
-    // tick on 1 second
-    endOfWinter.setTime(endOfWinter.getTime() + 1000);
-    message = datePipe.transform (endOfWinter.toISOString(), '+0100', 'dd MMMM yyyy HH:mm:ss.SSS');
-    expect(message).toBe('29 March 2020 02:00:00.000');
-  })
+  // test removed regarding time zone change at end of winter (change from 00:59 -> 02:00 removed as no longer relevant)
```
paul-graham, line 162: _Question:_
Again, why is this no longer relevant? We _do_ still have clocks going forwards and backwards... does the new component make it redundant? I'm unclear as to why it's not relevant. Can you explain further in this comment, please?

src/shared/components/palette/utils/date.pipe.ts
```diff
@@ -40,7 +40,8 @@ export class DatePipe implements PipeTransform {
         if (this.formatTrans && format && format !== 'short') {
           // support for java style formatting strings for dates
           format = this.translateDateFormat(format);
-          resultDate = formatDate(date, format, 'en-GB', zone)
+          const thisMoment = moment(date).format(format);
```
paul-graham, line 43: _Nitpick:_
Somewhat pointless assignment to a `const` and then immediate reassignment to `resultDate`.

src/shared/components/palette/utils/date.pipe.ts
```diff
@@ -67,7 +67,8 @@ export class DatePipe implements PipeTransform {
             return this.transform(shortISO, zone, format);
           }
           // If it did include time, we want a full ISO string.
-          return this.transform(d.toISOString(), zone, format);
+          const thisMoment = moment(d).format('YYYY-MM-DDTHH:mm:ss.SSSZ');
```
paul-graham, line 70: _Nitpick:_
Did you not previously declare a bunch of constants for date formats? It would be good to use one of those here rather than inline a string constant.

src/shared/domain/definition/case-field-model.spec.ts
```diff
@@ -49,6 +49,69 @@ describe('CaseField', () => {
...
+    component.display_context = null
+    expect(component.isMandatory()).toBe(false);
+    expect(component.isOptional()).toBe(false);
+    component.display_context = 'Mandatory';
+    expect(component.isMandatory()).toBe(true);
+    expect(component.isOptional()).toBe(false);
+    component.display_context = 'Optional';
+    expect(component.isMandatory()).toBe(false);
+    expect(component.isOptional()).toBe(true);
+    component.display_context = 'REAdOnlY';
```
paul-graham, line 64: _Question:_
Errr... 1337 speak? 🤔

src/shared/services/case-fields/format-translator.service.spec.ts
```diff
@@ -30,5 +30,48 @@ describe('FormatTranslatorService', () => {
       // tslint:disable-next-line:quotemark
       expect(formatTranslator.translate("dd MMM ''yyyy''")).toBe('dd MMM yyyy')
     })
+
+    it ('should remove time', () => {
+      expect(formatTranslator.removeTime('YYYY/MM/DD hh:mm:ss')).toBe('YYYY/MM/DD ::');
```
paul-graham, line 35: This output is a bit unpleasant... shouldn't it simply be `YYYY/MM/DD` without the extraneous ` ::`? 😬

src/shared/services/case-fields/format-translator.service.ts
```diff
@@ -169,4 +174,54 @@ export class FormatTranslatorService {
...
+      dateFormat = dateFormat.replace('h', '');
+    }
+    while (dateFormat.includes('m')) {
+      dateFormat = dateFormat.replace('m', '');
+    }
+    while (dateFormat.includes('S') || dateFormat.includes('s')) {
+      dateFormat = dateFormat.replace('S', '');
+      dateFormat = dateFormat.replace('s', '');
+    }
+    return dateFormat;
```
paul-graham, line 190: Yeah, as mentioned in the test, this doesn't seem to tidy up after itself when removing time. In the test example, it leaves ` ::` for a full `HH:mm:ss` timestamp. It'd leave ` :` for just `HH:mm` and it doesn't look like it handles milliseconds at all... so `HH:mm:ss.fff` (`10:19:28.123`) would leave you with ` ::.fff` (` ::.123`), which is clearly not what you want.

paul-graham (changes requested):
Some things that I definitely think need changing, some questions, and some nitpicks. I've tried to indicate appropriately in the comments which is which.

RELEASE-NOTES.md
```diff
@@ -1,5 +1,8 @@
 ## RELEASE NOTES
 
+### Version 2.79.15
```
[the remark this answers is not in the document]
connorpgpmcelroy: Will investigate after worked on other comments

package.json
```diff
@@ -43,26 +43,33 @@
...
     "minimist": "1.2.3",
     "mocha": "^5.0.5",
-    "moment": "^2.26.0",
+    "moment": "^2.29.1",
     "moment-timezone": "^0.5.31",
     "npm-registry-fetch": "^4.0.5",
     "rx-polling": "1.1.0",
     "typescript": "3.1.1",
     "underscore": "^1.9.1",
+    "xlsx": "^0.15.4",
```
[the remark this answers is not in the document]
connorpgpmcelroy: Will ask Sreekanth

src/shared/components/palette/date/read-date-field.component.ts
```diff
@@ -1,8 +1,18 @@
...
 import { AbstractFieldReadComponent } from '../base-field/abstract-field-read.component';
 
 @Component({
   selector: 'ccd-read-date-field',
   template: `<span class="text-16">{{caseField.value | ccdDate:'utc':caseField.dateTimeDisplayFormat}}</span>`
 })
-export class ReadDateFieldComponent extends AbstractFieldReadComponent {}
+export class ReadDateFieldComponent extends AbstractFieldReadComponent implements OnInit {
+  public timeZone = 'utc';
+/*   ngOnInit() {
```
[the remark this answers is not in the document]
connorpgpmcelroy: Yes, agreed and will get rid of it

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
```diff
@@ -0,0 +1,157 @@
...
+  encapsulation: ViewEncapsulation.None,
+  providers: [
+    {provide: NGX_MAT_DATE_FORMATS, useValue: CUSTOM_MOMENT_FORMATS},
+    {provide: NgxMatDateAdapter, useClass: NgxMatMomentAdapter},
+    {provide: NGX_MAT_MOMENT_DATE_ADAPTER_OPTIONS, useValue: {useUtc: true}}]
+})
+
+export class DatetimePickerComponent extends AbstractFormFieldComponent implements OnInit {
+
+  public disabled = false;
```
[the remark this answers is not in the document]
connorpgpmcelroy: I'll remove this because there was an expectation originally that the datetime picker might be disabled in some circumstances but now this is when the write date component is used

src/shared/components/palette/utils/date.pipe.spec.ts
connorpgpmcelroy: Will edit test instead of removing

src/shared/components/palette/utils/date.pipe.spec.ts
connorpgpmcelroy: Will explain here as well. Because of the use of moment library and therefore the fixed utc timezone of the datetime picker (that was enabled on instruction from Sreekanth), the read date component also needed to have a fixed utc timezone in order for the user to not get reasonably confused during the movement within pages. This means that any time given to either component will be used as the time and there will no longer be unnecessary movements between the time zones (i.e. giving one of the components a time that changes from what the user wants). This also fits in better with all use cases of the datepipe as there were occasions when timezones were changing twice (two hours behind or in front) when moving through the various steps.

src/shared/domain/definition/case-field-model.spec.ts
connorpgpmcelroy: This is testing the case changes in the isReadonly() as all letters should be set to the same case in order to equal 'READONLY' so I think it's beneficial. If you disagree, I can edit it to show in a more presentable format?

kuda1992 (changes requested):

src/shared/services/case-fields/format-translator.service.spec.ts
connorpgpmcelroy: I could put something in to remove colons when removing time from a datetime however there is no guarantee that colons will be used to separate the time. Could be 'YYYY:MM:DD hh-mm-ss' in some scenarios. I can make a reasonable assumption that this might be the case however - could just remove all colons if time removed?

src/shared/domain/definition/case-field-model.spec.ts
paul-graham: Okay, that's cool... maybe have an `expect(component.isReadonly()).toBe(false);` assertion below it too if it's case-sensitive. That would make it clear why it's there. :)

src/shared/services/case-fields/format-translator.service.ts
connorpgpmcelroy: Milliseconds are determined by character 'S' for datetime picker so they would be removed. Again, could make assumptions on formatting? Remove full stops and colons when removing time?

src/shared/services/case-fields/format-translator.service.spec.ts
paul-graham: I suppose you _could_ end up with a really stupid format like `'YYYY:hh-DD.mm MMM'` and then you'd be pretty screwed, although that's unlikely. There _is_ a way to sort this out once and for all (via regex) but it's probably more work than is justified. I think making some assumptions about what constitutes a sensible format is probably reasonable - i.e., `/`, `-`, ` `, and maybe `.` are reasonable separators for the date... so strip out anything else... and then do a trim on the whole string to get rid of any weird whitespace at the edges...?

demo/src/style/app.scss
```diff
@@ -228,7 +228,10 @@ select::-ms-expand {
   width: 100%;
 }
 .width-50 {
-  width: 50%;
+  width: 100%; // Overridden purely for demo purposes.
```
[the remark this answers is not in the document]
paul-graham: Because the actual component has a `width-50` class and is rendered that way within **Manage cases**. If I also stuck a `width-100` class on it, it might break it within **Manage cases** too. This is only within the demo, making it more visually appealing, and doesn't affect any production code so I thought it'd be fine. 🤷‍♂️

src/shared/components/palette/utils/date.pipe.spec.ts
paul-graham: As discussed, a comment to this effect would be awesome to prevent some future explorer stumbling into a moment-based timezone trap. 😁

src/shared/domain/definition/case-field-model.spec.ts
connorpgpmcelroy: No, the point is that the case is changed so that it's not case sensitive. A .toUpperCase() is used to get the case to be irrelevant

paul-graham: Ah, then `expect(component.isReadonly()).toBe(true);` for that one and not for the others! 😂

src/shared/services/case-fields/format-translator.service.spec.ts
connorpgpmcelroy: I'd rather not strip everything else out - I think that's a bit of an exclusive assumption and means any additional characters that we're not considering that the user might want are stripped out. However, definitely removing colons and trimming is valid, leaving an explanation in the comments stating why

package.json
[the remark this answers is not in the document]
connorpgpmcelroy: Traced this back to Adnan who I've asked for info

src/shared/components/palette/date/read-date-field.component.ts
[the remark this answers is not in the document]
connorpgpmcelroy: Removed

src/shared/components/palette/date/write-date-container.component.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+import { PaletteUtilsModule } from '../utils';
+import { WriteDateFieldComponent } from './write-date-field.component';
+import { CaseFieldService } from '../../../services';
+import { FormModule } from '../../../../components/form/form.module';
+import { WriteDateFieldContainerComponent } from './write-date-field-container.component';
+import { DatetimePickerComponent } from '../datetime-picker/datetime-picker.component';
+import { NgxMatDatetimePickerModule, NgxMatNativeDateModule, NgxMatTimepickerModule } from '@angular-material-components/datetime-picker';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { MatDatepickerModule, MatFormFieldModule, MatInputModule } from '@angular/material';
+import { FormatTranslatorService } from '../../../services/case-fields/format-translator.service';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Done

src/shared/components/palette/date/write-date-field-container.component.ts
```diff
@@ -0,0 +1,12 @@
+import { AbstractFormFieldComponent } from '../base-field/abstract-form-field.component';
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'ccd-write-date-container-field',
+  templateUrl: './write-date-container-field.html'
+})
+export class WriteDateFieldContainerComponent extends AbstractFormFieldComponent implements OnInit {
+
+  ngOnInit(): void {
```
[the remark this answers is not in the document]
connorpgpmcelroy: Yep, removed

src/shared/components/palette/datetime-picker/datetime-picker-component.spec.ts
```diff
@@ -0,0 +1,484 @@
...
+import { FormatTranslatorService } from '../../../services/case-fields/format-translator.service';
+import { DatetimePickerComponent } from './datetime-picker.component';
+import { By } from '@angular/platform-browser';
+import { NgxMatMomentAdapter } from '@angular-material-components/moment-adapter';
+import { NGX_MAT_DATE_FORMATS } from '@angular-material-components/datetime-picker';
+import { NgxMatDateAdapter } from '@angular-material-components/datetime-picker';
+import { CUSTOM_MOMENT_FORMATS } from './datetime-picker-utils';
+import { FieldLabelPipe, FirstErrorPipe } from '../utils';
+import { CaseFieldService } from '../../../services';
+import { CaseField, FieldType } from '../../../domain';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Done

src/shared/components/palette/datetime-picker/datetime-picker-component.spec.ts
```diff
@@ -0,0 +1,484 @@
...
+
+describe('DatetimePickerComponent', () => {
+
+  let component: DatetimePickerComponent;
+  let fixture: ComponentFixture<DatetimePickerComponent>;
+  let caseFieldService = new CaseFieldService();
+  // changing custom formats to make it easier to get date from input
+  CUSTOM_MOMENT_FORMATS.parse.dateInput = 'MM/DD/YYYY, HH:mm:SS';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Done

src/shared/components/palette/datetime-picker/datetime-picker.component.scss
```diff
@@ -0,0 +1,725 @@
+@import 'node_modules/govuk_frontend_toolkit/stylesheets/_shims';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Discussed - This is the current method of importing used so no change required

src/shared/components/palette/datetime-picker/datetime-picker.component.scss
```diff
@@ -0,0 +1,725 @@
...
+      outline: none;
+    }
+  }
+
+  .mat-icon-button {
+    height: 37px;
+    width: 37px;
+
+    &:focus {
+      outline: 1px solid orange;
```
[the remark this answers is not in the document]
connorpgpmcelroy: All 'oranges' have been changed

src/shared/components/palette/datetime-picker/datetime-picker.component.scss
```diff
@@ -0,0 +1,725 @@
...
+    }
+
+    // year selection panel
+    ngx-mat-multi-year-view {
+      .mat-calendar-table {
+        border-collapse: separate;
+        border-spacing: 10px;
+
+        .mat-calendar-body-active {
+          background-color: lightskyblue;
```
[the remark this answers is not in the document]
connorpgpmcelroy: Replaced with GDS 'light-blue'

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
```diff
@@ -0,0 +1,157 @@
...
+import { Component, ElementRef, Inject, Input, OnInit, ViewChild, ViewEncapsulation } from '@angular/core';
+import { FormControl } from '@angular/forms';
+import { MAT_DATE_LOCALE, ThemePalette } from '@angular/material';
+import { AbstractFormFieldComponent } from '../base-field/abstract-form-field.component';
+import { CaseField } from '../../../domain';
+import { NgxMatMomentAdapter, NGX_MAT_MOMENT_DATE_ADAPTER_OPTIONS } from '@angular-material-components/moment-adapter';
+import { CUSTOM_MOMENT_FORMATS } from './datetime-picker-utils';
+import { FormatTranslatorService } from '../../../services/case-fields/format-translator.service';
+import { Moment } from 'moment/moment';
+import moment = require('moment/moment');
```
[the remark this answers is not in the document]
connorpgpmcelroy: Fixed

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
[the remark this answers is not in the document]
connorpgpmcelroy: Removed

src/shared/components/palette/dynamic-list/write-dynamic-list-field.component.ts
```diff
@@ -22,10 +22,9 @@ export class WriteDynamicListFieldComponent extends AbstractFieldWriteComponent
     /**
      * Reassigning value from formatted_value when value is empty
      */
-    if (!this.caseField.value) {
-      if (this.caseField.formatted_value && this.caseField.formatted_value.value && this.caseField.formatted_value.value.code) {
-        this.caseField.value = this.caseField.formatted_value.value.code;
-      }
+    if (!this.caseField.value && this.caseField.formatted_value
```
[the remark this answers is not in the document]
connorpgpmcelroy: I didn't make this change but I've changed it back (I had to separate the long if statement before because linting was failing)

src/shared/components/palette/palette.module.ts
```diff
@@ -50,6 +53,7 @@ import { FileUploadProgressGuard } from './document/file-upload-progress.guard';
 import { FileUploadStateService } from './document/file-upload-state.service';
 import { OrganisationModule } from './organisation/organisation.module';
 import { ReadOrganisationFieldComponent, WriteOrganisationFieldComponent } from './organisation';
+import { WriteDateFieldContainerComponent } from './date';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Done

src/shared/components/palette/palette.service.ts
```diff
@@ -40,6 +39,7 @@ import { WriteFixedRadioListFieldComponent, ReadFixedRadioListFieldComponent } f
 import { CaseHistoryViewerFieldComponent } from './history';
 import { WriteOrganisationFieldComponent } from './organisation/write-organisation-field.component';
 import { ReadOrganisationFieldComponent } from './organisation/read-organisation-field.component';
+import { WriteDateFieldContainerComponent } from './date/write-date-field-container.component';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Sorted

src/shared/components/palette/utils/date.pipe.spec.ts
connorpgpmcelroy: Done

src/shared/components/palette/utils/date.pipe.spec.ts
connorpgpmcelroy: Done

src/shared/components/palette/utils/date.pipe.ts
connorpgpmcelroy: Done

src/shared/components/palette/utils/date.pipe.ts
connorpgpmcelroy: Not completely sure what the suggestion was here actually but I have used a constant for the format and placed it at the start of the translate method for use within it

src/shared/domain/definition/case-field-model.spec.ts
connorpgpmcelroy: There is a test for it above (I've changed the mention of it here to not be 1337)

src/shared/services/case-fields/format-translator.service.spec.ts
connorpgpmcelroy: Change made

src/shared/services/case-fields/format-translator.service.ts
connorpgpmcelroy: Changes made as per above

demo/src/style/app.scss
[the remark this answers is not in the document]
connorpgpmcelroy: Unless there's disagreement I'm treating this as resolved (although leaving this open so you can disagree if not)

demo/test/mockData/searchResult.js
```diff
@@ -0,0 +1,22 @@
...
+    total: 0
+}
+
+function getDatetimefieldsResults(){
+    const searchResults = {
+        columns: [],
+        results: [],
+        total: 0
+    }
+
```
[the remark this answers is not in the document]
connorpgpmcelroy: Removed

demo/test/mockData/workbasketSearchInput.js
```diff
@@ -0,0 +1,30 @@
...
+
+function getMockJurisdictionWorkbaseketConfig() {
+    const workBasketConfig = new CCDWorkbasketConfig();
+    workBasketConfig
+        .addField({ id: "simpletext", type: "Text", label: "Simple text input" })
+        .addField({ id: "radioInput", type: "FixedRadioList", label: "Simple Radio input", list: [{ code: "a", label: "A" }, { code: "b", label: "B" }, { code: "c", label: "C" }] })
+        .addField({ id: "radioYesorNo", type: "YesOrNo", label: "Simple Yes or No input" })
+        .addField({ id: "fixedListItem", type: "FixedList", label: "fixed listinput", list: [{ code: "a", label: "A" }, { code: "b", label: "B" }, { code: "c", label: "C" }] })
+        .addField({ id: "multiSelectItem", type: "MultiSelectList", label: "Multi select input", list: [{ code: "a", label: "A" }, { code: "b", label: "B" }, { code: "c", label: "C" }] })
+
```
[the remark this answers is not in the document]
connorpgpmcelroy: Fixed whitespace (removed any double lines)

src/shared/components/palette/date/read-date-field.component.ts
```diff
@@ -1,8 +1,18 @@
...
+import { Component, OnInit } from '@angular/core';
 import { AbstractFieldReadComponent } from '../base-field/abstract-field-read.component';
 
 @Component({
   selector: 'ccd-read-date-field',
   template: `<span class="text-16">{{caseField.value | ccdDate:'utc':caseField.dateTimeDisplayFormat}}</span>`
 })
-export class ReadDateFieldComponent extends AbstractFieldReadComponent {}
+export class ReadDateFieldComponent extends AbstractFieldReadComponent implements OnInit {
+  public timeZone = 'utc';
```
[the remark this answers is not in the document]
connorpgpmcelroy: Already done

src/shared/components/palette/date/write-date-field-container.component.ts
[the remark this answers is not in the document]
connorpgpmcelroy: Done already

package.json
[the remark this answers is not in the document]
connorpgpmcelroy: Adnan has said unlikely to affect anything if replaced as devDependency. Have tested by running yarn with package.json and then running the demo and did not flag up anything as an issue

RELEASE-NOTES.md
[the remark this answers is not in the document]
connorpgpmcelroy: Spoke to Mo about this and he said 'it needs to be master version you referenced because the the pre-tagged is currently being tested'. Mo said they are using 'tagged pre-release version whilst the master candidate is being regression tested so for the devs its: 3.0.0-rc.1' but it's not relevant to the merge

src/shared/components/palette/date/read-date-field.component.ts
```diff
@@ -1,8 +1,10 @@
-import { Component } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
```
paul-graham, line 1: _Nitpick:_
`OnInit` no longer needs to be imported as as the component doesn't implement the interface.

src/shared/components/palette/date/write-date-container-field.component.ts
```diff
@@ -0,0 +1,8 @@
+import { AbstractFormFieldComponent } from '../base-field/abstract-form-field.component';
+import { Component, OnInit } from '@angular/core';
```
paul-graham, line 2: _Nitpick:_
Same here - we don't need `OnInit` as the component doesn't implement the interface.

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
```diff
@@ -0,0 +1,156 @@
...
+  @ViewChild('input') inputElement: ElementRef<HTMLInputElement>;
+  private dateTimeEntryFormat;
+
+  constructor(private readonly formatTranslationService: FormatTranslatorService,
+              @Inject(NGX_MAT_DATE_FORMATS) private ngxMatDateFormats: NgxMatDateFormats) {
+
+    super();
+  }
+
+  ngOnInit(): void {
```
paul-graham, line 58: _Nitpick:_
We agreed at the last Tech Forum that these will always be `public ngOnInit(): void` going forward.

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
```diff
@@ -0,0 +1,156 @@
...
+    this.setDateTimeFormat();
+    this.dateControl = this.registerControl(new FormControl(this.caseField.value)) as FormControl;
+    // in resetting the format just after the page initialises, the input can be reformatted
+    // otherwise the last format given will be how the text shown will be displayed
+    setTimeout(() => {
+      this.setDateTimeFormat();
+    }, 1);
+  }
+
+  setDateTimeFormat(): void {
```
paul-graham, line 70: _Nitpick:_
Following on from the above, it was also agreed that `public` methods would be explicitly declared as such...

src/shared/components/palette/utils/date.pipe.spec.ts
```diff
@@ -54,8 +54,6 @@ describe('DatePipe', () => {
   it('should render correct date if UTC date in yyyy-mm-ddThh:mm:ss PM format', () => {
     const firstPass: string = datePipe.transform('2017-07-26T20:10:05', 'local', null);
     expect(firstPass).toBe('26 Jul 2017, ' + getExpectedHour(8) + ':10:05 PM');
-    const secondPass: string = datePipe.transform(firstPass, 'local', null);
```
paul-graham, line 57: I added this test in to make sure the pipe doesn't break if it gets a date a second time. This was because there were circumstances on the Case details tabs where dates were being parsed twice and weren't rendering correctly.

If this has been removed because it's failing, that will regress the fix.

paul-graham (changes requested):
Bit concerned about one of the unit tests having assertions removed. If the test is failing now, the `DatePipe` has been broken, which will cause a regression on the Case details view.

src/shared/components/palette/date/read-date-field.component.ts
connorpgpmcelroy: Done

src/shared/components/palette/date/write-date-container-field.component.ts
connorpgpmcelroy: Done

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
connorpgpmcelroy: Done

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
connorpgpmcelroy: Done

paul-graham (approved):
Thanks for making all the requested changes, @connorpgpmcelroy.

This _can't_ count as one of the 2 approvals but there's no other way for me to remove the "Request changes" thing.

leepavlouhmcts (approved):

udaydenduluri33 (approved):
LGTMc

PR #786, opened 2021-05-18
Bugfix/e UI 2792 datetime picker integration

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-2792

### Change description ###
Fix two issues:
Stop day from appearing instead of date
Return valid value instead of appended 'Z' value

```
[ ] Yes
[  x ] No
```


udaydenduluri33 (approved):
LGTMc

leepavlouhmcts (approved):

PR #791, opened 2021-05-25
Make pcm role check

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-4077

### Change description ###
Checking for pui-case-manager role to determine whether searchForCompletable needs to be called

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/shared/components/case-editor/services/cases.service.ts
+++ b/src/shared/components/case-editor/services/cases.service.ts
@@ -41,6 +41,8 @@ export class CasesService {
   public static readonly SERVER_RESPONSE_FIELD_TYPE_COMPLEX = 'Complex';
   public static readonly SERVER_RESPONSE_FIELD_TYPE_DYNAMIC_LIST = 'DynamicList';
 
+  public static readonly PUI_CASE_MANAGER = 'pui-case-manager';
+
   /**
    *
    * @type {(caseId:string)=>"../../Observable".Observable<Case>}
@@ -353,7 +355,7 @@ export class CasesService {
     const userInfoStr = this.sessionStorageService.getItem('userDetails');
     if (userInfoStr) {
       const userInfo: UserInfo = JSON.parse(userInfoStr);
-      if (userInfo && userInfo.roles.indexOf('pui-case-manager') !== -1) {
+      if (userInfo && userInfo.roles && userInfo.roles.indexOf(CasesService.PUI_CASE_MANAGER) !== -1) {
         return true;
       }
     }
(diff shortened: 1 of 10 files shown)


kuda1992 (approved):
looks good

src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -343,4 +345,18 @@ export class CasesService {
...
   }
+
+  /*
+  Checks if the user has role of pui-case-manager and returns true or false
+  */
+  private isPuiCaseManager(): boolean {
+    const userInfoStr = this.sessionStorageService.getItem('userDetails');
+    if (userInfoStr) {
+      const userInfo: UserInfo = JSON.parse(userInfoStr);
+      if (userInfo && userInfo.roles.indexOf('pui-case-manager') !== -1) {
```
udaydenduluri33, line 356: please add a check to see if userInfo.roles is present.

udaydenduluri33: Also, move 'pui-case-manager' to a public constant

connorpgpmcelroy: ok, will do

udaydenduluri33 (dismissed):
LGTMc

udaydenduluri33 (dismissed):
LGTMc

src/shared/components/case-editor/services/cases.service.ts
```diff
@@ -343,4 +347,16 @@ export class CasesService {
...
   }
+
+  /*
+  Checks if the user has role of pui-case-manager and returns true or false
+  */
+  private isPuiCaseManager(): boolean {
+    const userInfoStr = this.sessionStorageService.getItem('userDetails');
+    if (userInfoStr) {
+      const userInfo: UserInfo = JSON.parse(userInfoStr);
+      return userInfo && userInfo.roles && userInfo.roles.indexOf(CasesService.PUI_CASE_MANAGER) !== -1
```
andywilkinshmcts, line 358: This line is dodgy - logical && has higher precedence that strict inequality, so I think this means 
(userInfo && userInfo.roles && userInfo.roles.indexOf(CasesService.PUI_CASE_MANAGER)) !== -1 when it should be 
userInfo && userInfo.roles && (userInfo.roles.indexOf(CasesService.PUI_CASE_MANAGER) !== -1)
Since nobody can ever remember the precedence table, it's always best to have the brackets

andywilkinshmcts (changes requested):
Need to fix the brackets in isPuiCaseManager

src/shared/components/case-editor/services/cases.service.ts
connorpgpmcelroy: ok, good point, will rectify

andywilkinshmcts (approved):

leepavlouhmcts (approved):

udaydenduluri33 (approved):

PR #802, opened 2021-06-08
Bugfix/EUI-3943 Nested collection is not rendering correctly on case type tab for table view

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-3943?filter=-1

### Change description ###
Remove the `ccdConditionalShow` Directive and use `ReadFieldsFilterPipe`

```
[x] Yes
[] No
```


RELEASE-NOTES.md
```diff
@@ -11,6 +14,7 @@
 **EUI-3155** Accessibility amends for non-descriptive headings
 **EUI-3550** Replace heading tags with spans on case edit pages
 
+
```
paul-graham, line 17: _Nitpick:_
You might be okay in a `.md` file but lint complains about multiple blank lines of spacing... and this seems like an accidental addition.

src/case-ui-toolkit.module.ts
```diff
@@ -108,7 +108,6 @@ import { CaseHistoryModule } from './shared/components/case-history';
       WriteEmailFieldComponent,
       WriteDateFieldComponent,
 
-      ConditionalShowDirective,
```
paul-graham, line 111: I may be wrong, but I think removing this would constitute a breaking change. MC is not the only consumer of the CCD Case UI Toolkit and if anything else is leveraging that, it'll break. It's also not needed for the change you're making.

src/shared/components/palette/complex/ccd-collection-table-value-case-fields.pipe.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { CcdCollectionTableCaseFieldsFilterPipe } from './ccd-collection-table-value-case-fields.pipe';
+
+describe('CcdCollectionTableCaseFieldsFilterPipe', () => {
+  it('create an instance', () => {
+    const pipe = new CcdCollectionTableCaseFieldsFilterPipe();
+    expect(pipe).toBeTruthy();
```
paul-graham, line 6: This doesn't test anything. 🤷‍♂️

src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
```diff
@@ -154,15 +120,51 @@ export class ReadComplexFieldCollectionTableComponent extends AbstractFieldReadC
...
+      } else {
+        labelsVertical[obj.id] = {label: obj.label, type: {type: obj.field_type.type}, caseField: obj};
+        allLabels[obj.id] = {label: obj.label, type: {type: obj.field_type.type}, caseField: obj};
+      }
+    }
   }
 
-  sortWidget(column: any) {
-    return this.isSortAscending(column) ? '&#9660;' : '&#9650;';
+  private populateCaseFieldValuesIntoRows() {
```
paul-graham, line 154: _Nitpick:_
Should have a `: void` return type.

src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
```diff
@@ -154,15 +120,51 @@ export class ReadComplexFieldCollectionTableComponent extends AbstractFieldReadC
...
+    for (let obj of this.caseField.value) {
+      this.rows.push(obj.value);
+      this.isHidden.push(true);
+    }
   }
 
-  trackByIndex(index: number, obj: any): any {
-    return index;
+  private isVerticleDataNotEmpty(row) {
```
paul-graham, line 161: _Nitpick:_
Should have a : void return type.

src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
```diff
@@ -154,15 +120,51 @@ export class ReadComplexFieldCollectionTableComponent extends AbstractFieldReadC
...
+                                   allLabels: { [p: string]: any },
+                                   labelsVertical: { [p: string]: any }) {
+    for (let id of this.columns) {
+      labelsHorizontal[id.trim()] = allLabels[id.trim()];
+      labelsHorizontal[id.trim()].sortOrder = SortOrder.UNSORTED;
+      delete labelsVertical[id.trim()];
+    }
+  }
+
+  private populateLabels(labelsVertical: { [p: string]: any }, allLabels: { [p: string]: any }) {
```
paul-graham, line 137: _Nitpick:_
Should have a : void return type.

src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
```diff
@@ -154,15 +120,51 @@ export class ReadComplexFieldCollectionTableComponent extends AbstractFieldReadC
...
-    return !(column.sortOrder === SortOrder.UNSORTED || column.sortOrder === SortOrder.DESCENDING);
+  public sortWidget(column: any): string {
+    return ReadComplexFieldCollectionTableComponent.isSortAscending(column) ? '&#9660;' : '&#9650;';
+  }
+
+  private populateHorizontalLabels(labelsHorizontal: { [p: string]: any },
+                                   allLabels: { [p: string]: any },
+                                   labelsVertical: { [p: string]: any }) {
+    for (let id of this.columns) {
+      labelsHorizontal[id.trim()] = allLabels[id.trim()];
```
paul-graham, line 131: _Nitpick:_
Why do `id.trim()` on every single instance it's being used rather than `const trimId: string = id.trim();` in the line above?

src/shared/directives/conditional-show/conditional-show.directive.ts
```diff
@@ -1,213 +0,0 @@
-import { AfterViewInit, Directive, ElementRef, Input, OnDestroy, Renderer2 } from '@angular/core';
```
paul-graham, line 1: Why has this been removed as part of this PR? Wasn't it removed in a previous PR? This feels like a risky thing to do as part of this PR as it's not absolutely necessary... or is it? 😬

paul-graham (changes requested):
There might be some linting issues with this, particularly around import ordering, but there are some things I think need changing aside from that and I'm concerned that this includes the removal of the `ConditionalShowDirective` when it doesn't seem to be needed.

RELEASE-NOTES.md
kuda1992: Addressed

src/shared/components/palette/complex/ccd-collection-table-value-case-fields.pipe.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+      complex_fields: [],
+      id: 'Hearing',
+      type: 'Complex'
+    },
+    display_context_parameter: '#TABLE(Hearing_type,Hearing_venue)'
+  }
+  it('to create complex case field with a single case field', () => {
+    const pipe = new CcdCollectionTableCaseFieldsFilterPipe();
+    const RESULT = pipe.transform(CASE_FIELDS, CASE_FIELD as unknown as CaseField, {})
+    expect(RESULT.field_type.complex_fields.length).toBe(1);
```
paul-graham, line 40: Still not really sure what value there is with this test... and I'm also not sure why you're doing `CASE_FIELD as unknown as CaseField`. What does the `as unknown` bit do in the context of this test?

src/shared/components/palette/complex/read-complex-field-collection-table.component.ts
```diff
@@ -90,33 +60,29 @@ export class ReadComplexFieldCollectionTableComponent extends AbstractFieldReadC
...
-  }
-
-  isNotBlank(value: string) {
-    return value !== null && value !== '';
+    return true
   }
 
-  addCaseReferenceValue(field, value: any) {
-    field.value = { CaseReference: value};
+  public addCaseReferenceValue(field, value: any) {
```
paul-graham, line 68: Should have a return type. Must've missed this the first time...

paul-graham (changes requested):
I thought the agreement was that this was going up by a major version...? And I'm still not sure the single unit test for the pipe is useful. It doesn't really _test_ anything.

paul-graham (approved):
Much better set of unit tests for the new pipe. Thanks for making the requested changes, @kuda1992. :)

phillip-whitaker (approved):
LGBTQ+

PR #804, opened 2021-06-09
defaulted hidden state for added complex fields in collection to false

phillip-whitaker:

https://tools.hmcts.net/jira/browse/EUI-3983

#### The Issue
When a collection that contains complex fields is added as part of a collection the fields with show condition are momentarily displayed before the `conditional-show-form.directive.ts` kicks in and hides it. This is because the subscription has a debounce(100) on it. The debounce is there to help performance.

#### The Fix
As I saw it, there are two fixes to this bug, remove the debounce or default added collection complex fields to hidden = true for fields that have show_condition logic. I chose the latter because I think is more sustainable and I know that the show/hide logic all being changed soon too so this fix will withstand the changes to the current directive.

```
[ ] Yes
[x-x-x] No
```

### DIFF
--- a/src/shared/components/palette/collection/write-collection-field.component.ts
+++ b/src/shared/components/palette/collection/write-collection-field.component.ts
@@ -160,7 +160,7 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
     const fieldType = plainToClassFromExist(new FieldType(), this.caseField.field_type.collection_field_type);
     if (fieldType.complex_fields) {
       fieldType.complex_fields
-        .filter((cf: CaseField) => !cf.show_condition)
+        .filter((cf: CaseField) => !!cf.show_condition)
         .map((cf: CaseField) => cf.hidden = true);
     }
 
--- a/src/shared/components/palette/collection/write-collection-field.component.ts
+++ b/src/shared/components/palette/collection/write-collection-field.component.ts
@@ -2,11 +2,11 @@ import { Component, ElementRef, Input, OnDestroy, OnInit, QueryList, ViewChildre
 import { FormArray, FormControl, FormGroup } from '@angular/forms';
 import { MatDialog, MatDialogConfig } from '@angular/material';
 import { ScrollToService } from '@nicky-lenaers/ngx-scroll-to';
-import { plainToClassFromExist } from 'class-transformer';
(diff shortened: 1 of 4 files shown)


Mo-Lala84 (approved):
lgtm

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -156,11 +156,19 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
 
   private newCaseField(id: string, item, index, isNew = false) {
     const isNotAuthorisedToUpdate = !isNew && this.isNotAuthorisedToUpdate(index);
+
+    const fieldType = plainToClassFromExist(new FieldType(), this.caseField.field_type.collection_field_type);
+    if (fieldType.complex_fields) {
+      fieldType.complex_fields
+        .filter((cf: CaseField) => cf.show_condition !== null)
```
paul-graham, line 163: Do we know for sure this is always `null` when it's not set? It's never `undefined` or `''`? `!!cf.show_condition` would account for all of those, which might be a better option.

paul-graham (commented):
Just a question around the filtering mechanism...

src/shared/components/palette/collection/write-collection-field.component.ts
```diff
@@ -156,11 +156,19 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
 
   private newCaseField(id: string, item, index, isNew = false) {
     const isNotAuthorisedToUpdate = !isNew && this.isNotAuthorisedToUpdate(index);
+
+    const fieldType = plainToClassFromExist(new FieldType(), this.caseField.field_type.collection_field_type);
+    if (fieldType.complex_fields) {
+      fieldType.complex_fields
+        .filter((cf: CaseField) => !cf.show_condition)
```
paul-graham, line 163: This should be a double bang, not a single. You're filtering out anything _with_ a `show_condition` now! 😂

paul-graham (changes requested):

paul-graham (approved):
Sorted. Thanks for fixing the double bang, @phillip-whitaker.

paul-graham (dismissed):

paul-graham (approved):

PR #828, opened 2021-06-21
Amend removeItem to ensure proper control removal

paulhowes-HMCTS:
https://tools.hmcts.net/jira/browse/EUI-4072

### Change description ###
Amend the removeItem method to prevent subsequent form collection elements do not get removed when removing a preceding collection element.

```
[ ] Yes
[x] No
```


src/shared/components/palette/collection/write-collection-field.component.spec.ts
```diff
@@ -269,8 +269,6 @@ describe('WriteCollectionFieldComponent', () => {
     let writeFields = de.queryAll($WRITE_FIELDS);
     expect(writeFields.length).toBe(VALUES.length - 1);
 
-    let field2 = writeFields[0].componentInstance;
```
paulhowes-HMCTS, line 272: After discussion with @Mo-Lala84 it was agreed that this would be removed. Further information can be found on the corresponding Jira ticket.

paul-graham (dismissed):

leepavlouhmcts (approved):

paul-graham (approved):
Previously approved this and all that's changed here is the version number... so all looks good from my point of view. 🙂

PR #833, opened 2021-06-23
Bugfix/EUI-4213 no doesn't enable continue button when yes no in complex object

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-4213

### Change description ###
Hidden complex types are preventing users from progressing further to fix the issue `PageValidationService.isHidden` will check for complex fields and strip out hidden fields. 

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/shared/components/case-editor/services/page-validation.service.ts
+++ b/src/shared/components/case-editor/services/page-validation.service.ts
@@ -8,16 +8,26 @@ import { WizardPage } from '../domain/wizard-page.model';
 
 @Injectable()
 export class PageValidationService {
-  constructor(private caseFieldService: CaseFieldService) {}
+  constructor(private caseFieldService: CaseFieldService) {
+  }
 
   isPageValid(page: WizardPage, editForm: FormGroup): boolean {
     return page.case_fields
       .filter(caseField => !this.caseFieldService.isReadOnly(caseField))
-      .filter(caseField => !this.isHidden(caseField, editForm))
+      .filter(caseField => this.isHidden(caseField, editForm))
       .every(caseField => {
         let theControl = editForm.controls['data'].get(caseField.id);
         return this.checkDocumentField(caseField, theControl) && this.checkOptionalField(caseField, theControl);
       });
+  }
+
+  public isHidden(caseField: CaseField, editForm: FormGroup): boolean {
+    const formFields = editForm.getRawValue();
+    if (caseField.field_type.complex_fields && caseField.field_type.complex_fields.length) {
+      return !caseField.field_type.complex_fields.some((cf: CaseField) => this.isHidden(cf, editForm))
+    }
+    const condition = ShowCondition.getInstance(caseField.show_condition);
+    return condition.match(formFields.data);
   }
 
   private checkDocumentField(caseField: CaseField, theControl: AbstractControl): boolean {
@@ -25,12 +35,6 @@ export class PageValidationService {
       return true;
     }
     return !(this.checkMandatoryField(caseField, theControl));
-  }
-
-  public isHidden(caseField: CaseField, editForm: FormGroup): boolean {
-    const formFields = editForm.getRawValue();
-    const condition = ShowCondition.getInstance(caseField.show_condition);
-    return !condition.match(formFields.data);
   }
 
   private checkOptionalField(caseField: CaseField, theControl: AbstractControl): boolean {
--- a/src/shared/components/case-editor/services/page-validation.service.ts
+++ b/src/shared/components/case-editor/services/page-validation.service.ts
@@ -14,7 +14,7 @@ export class PageValidationService {
   isPageValid(page: WizardPage, editForm: FormGroup): boolean {
     return page.case_fields
       .filter(caseField => !this.caseFieldService.isReadOnly(caseField))
-      .filter(caseField => this.isHidden(caseField, editForm))
+      .filter(caseField => !this.isHidden(caseField, editForm))
       .every(caseField => {
         let theControl = editForm.controls['data'].get(caseField.id);
         return this.checkDocumentField(caseField, theControl) && this.checkOptionalField(caseField, theControl);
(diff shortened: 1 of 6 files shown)


src/shared/components/case-editor/services/page-validation.service.ts
```diff
@@ -8,31 +8,35 @@ import { WizardPage } from '../domain/wizard-page.model';
...
       .filter(caseField => !this.caseFieldService.isReadOnly(caseField))
-      .filter(caseField => !this.isHidden(caseField, editForm))
+      .filter(caseField => this.isHidden(caseField, editForm))
       .every(caseField => {
         let theControl = editForm.controls['data'].get(caseField.id);
         return this.checkDocumentField(caseField, theControl) && this.checkOptionalField(caseField, theControl);
       });
   }
 
+  public isHidden(caseField: CaseField, editForm: FormGroup): boolean {
```
paul-graham, line 24: This is _not_ indicating when a field is hidden - if it matches the `show_condition` it is, by definition, shown. Can you rename this method, please, or else it'll be confusing... especially seeing as you've flipped the logic on line 17 so it's apparently now filtering out _shown_ fields.

paul-graham (changes requested):

src/shared/components/case-editor/services/page-validation.service.ts
kuda1992: I will revert the logic since it won't make sense in other contexts the method is being used. Thanks for pointing this out.

src/shared/components/case-editor/services/page-validation.service.ts
```diff
@@ -20,19 +21,22 @@ export class PageValidationService {
...
+    const formFields = editForm.getRawValue();
+    if (caseField.field_type.complex_fields && caseField.field_type.complex_fields.length) {
+      return caseField.field_type.complex_fields.some((cf: CaseField) => !this.isHidden(cf, editForm))
```
paul-graham, line 27: Trying to understand this bit of logic now... but needs a semicolon on the end either way.
`return caseField.field_type.complex_fields.some((cf: CaseField) => !this.isHidden(cf, editForm));`

`complex_fields => [ shown_field, shown_field ]`
=> `.some(...) => true`, which says that `caseField` _is_ hidden, even though all of the fields it contains are shown?
`complex_fields => [ hidden_field, hidden_field_2 ]`
=> `.some(...) => false`, which says that `caseField` _isn't_ hidden, even though all of the fields it contains _are_ hidden...?

Maybe it should be this?
`!return caseField.field_type.complex_fields.some((cf: CaseField) => !this.isHidden(cf, editForm));`

`complex_fields => [ shown_field, shown_field ]`
=> `!.some(...) => false`, which says that `caseField` _isn't_ hidden, which is what we want.
`complex_fields => [ hidden_field, hidden_field_2 ]`
=> `!.some(...) => true`, which says that `caseField` _is_ hidden since none of the fields within it are shown, which is also what we want.

Or have I misunderstood the intent here? 😬

paul-graham (changes requested):
Still unsure about the logic within `isHidden(...)`... 🤔

src/shared/components/case-editor/services/page-validation.service.ts
kuda1992: @paul-graham The logic should be if we have some hidden fields do not bother checking the validity of the fields inside that complex.
`if (caseField.field_type.complex_fields && caseField.field_type.complex_fields.length) {
      const hiddenFields = caseField.field_type.complex_fields.filter((cf: CaseField) => this.isHidden(cf, editForm))
      return !hiddenFields.length
    }`

paul-graham: But that still doesn't make any sense to me, which is why I'm struggling to understand what you're trying to do with this. This is what it appears to be doing (in pseudocode):

```
function isHidden(field: CaseField): boolean {
  if (caseField.isComplex()) {
    return caseField.hasNoHiddenSubFields(); // 1. This makes no sense to me. It seems the opposite of what we want.
  }
  return caseField.isHidden(); // 2. This bit makes sense.
  // 3. Returning at 1. for ALL complex fields also removes the possibility of the complex field
  //    ITSELF being hidden... is that not possible?
}
```

Non-complex fields:
1. CaseField is HIDDEN
** `isHidden(...) => true` ✅
2. CaseField is SHOWN
** `isHidden(...) => false` ✅

Complex fields (2 sub-fields, A and B):
1. CaseField is HIDDEN (A is SHOWN, B is SHOWN)
** `isHidden(...) => true` ✅
2. CaseField is HIDDEN (A is HIDDEN, B is SHOWN)
** `isHidden(...) => false` ❌
3. CaseField is HIDDEN (A is SHOWN, B is HIDDEN)
** `isHidden(...) => false` ❌
4. CaseField is HIDDEN (A is HIDDEN, B is HIDDEN)
** `isHidden(...) => false` ❌
5. CaseField is SHOWN (A is SHOWN, B is SHOWN)
** `isHidden(...) => true` ❌
6. CaseField is SHOWN (A is HIDDEN, B is SHOWN)
** `isHidden(...) => false` ✅
7. CaseField is SHOWN (A is SHOWN, B is HIDDEN)
** `isHidden(...) => false` ✅
8. CaseField is SHOWN (A is HIDDEN, B is HIDDEN)
** `isHidden(...) => false` ✅

I _think_ the logic should be something more like this:
```
public isHidden(caseField: CaseField, editForm: FormGroup): boolean {
  const formFields = editForm.getRawValue();
  if (caseField.field_type.complex_fields && caseField.field_type.complex_fields.length > 0) {
    // Check to see if ALL of the sub-fields are hidden.
    const allHidden = caseField.field_type.complex_fields.every((cf: CaseField) => this.isHidden(cf, editForm));

    // If they're ALL hidden, this complex field should be considered hidden too so return early.
    if (allHidden) {
      return true;
    }
  }

  // If we get here, either it's not a complex field or it IS a complex field and has at least one visible sub-field
  // so we need to see whether the complex field ITSELF is hidden.
  const condition = ShowCondition.getInstance(caseField.show_condition);
  return !condition.match(formFields.data);
}
```

paul-graham (dismissed):
Nice. Thanks for walking me through the issue you were trying to solve, @kuda1992. This looks like it'll do the job. 😁

PR #836, opened 2021-06-28
EUI-4107-Document-type

ishOsw:
- [x] README and other documentation has been updated / added (if needed)

https://tools.hmcts.net/jira/browse/EUI-4107

### Change description ###

Extend document types to pass into the MediaViewer

```
[ ] Yes
[x] No
```


src/shared/services/document-management/document-management.service.ts
```diff
@@ -50,14 +58,35 @@ export class DocumentManagementService {
       if (position === documentFieldValue.document_filename.length) {
         fileExtension = '';
       } else if (position >= 0) {
-        fileExtension = documentFieldValue.document_filename.slice(position + 1);
+        fileExtension = documentFieldValue.document_filename.slice(position + 1).toUpperCase();
```
adnanakgun, line 61: Hey @ishOsw,

The `.toUpperCase()` could be what is breaking this. In any case this is breaking the test. You need to come up with a solution that doesn't break what is already there.

adnanakgun: The test for `should return media viewer data for tif content types` in src/shared/services/document-management/document-management.service.spec.ts  specifically deals with the issue of case sensitivity.

Mo-Lala84 (dismissed):
approved

src/shared/services/document-management/document-management.service.ts
```diff
@@ -55,16 +63,37 @@ export class DocumentManagementService {
...
+      return DocumentManagementService.RTF;
     } else if (fileExtension.toLowerCase() === 'pdf') {
       return DocumentManagementService.PDF;
     } else {
-      console.warn(`Unknown content type with the file extension: ${fileExtension}`);
       return fileExtension;
     }
   }
 
   isImage(imageType: string) {
```
paul-graham, line 83: _Nitpick:_
Should have a return type of `boolean` and a `public`/`private` indicator. The same for all of the `isX` ones in here, actually...

paul-graham (commented):
Separate discussion about removing the CSV type, which isn't supported by the API... but also this little nitpick if you have time for it. 🙂

src/shared/services/document-management/document-management.service.ts
adnanakgun: You are right of course.

paul-graham (approved):
Excellent stuff. Thank you for making the requested changes, @adnanakgun, including my little nitpick!

Mo-Lala84 (approved):
approved

PR #852, opened 2021-07-05
EUI-4244: Mid-event callback error

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-4244

### Change description ###
Initial fix (identified by @RiteshHMCTS) was to appropriately clear out complex objects that have null values when posting mid-event callbacks. Not doing this caused problems for "optional" fields that were mandatory based on some business logic within CCD.

This then presented with a different issue in `release-4.1.0` as a result of getting label interpolation working for multi-selects. In order to sorted this, added a method to clear out properties that are only there for the multi-select labels when submitting data, either as part of a mid-event callback or at the end of the flow after Check Your Answers.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/services/fields/fields.utils.ts
+++ b/src/shared/services/fields/fields.utils.ts
@@ -15,7 +15,8 @@ export class FieldsUtils {
 
   private static readonly currencyPipe: CurrencyPipe = new CurrencyPipe('en-GB');
   private static readonly datePipe: DatePipe = new DatePipe(new FormatTranslatorService());
-  public static readonly LABEL_SUFFIX = '-LABEL';
+  // EUI-4244. 3 dashes instead of 1 to make this less likely to clash with a real field.
+  public static readonly LABEL_SUFFIX = '---LABEL';
 
   public static convertToCaseField(obj: any): CaseField {
     if (!(obj instanceof CaseField)) {
--- a/src/shared/services/form/form-value.service.ts
+++ b/src/shared/services/form/form-value.service.ts
@@ -9,7 +9,7 @@ export class FormValueService {
   /**
    * Gets value of a field based on fieldKey which is a dot separated reference to value and collection index.
    * There are two exeptions:
-   * 1) In case of a multiselect being identified as a leaf a '---LABEL' suffix is appended to the key and values og that key are returned
+   * 1) In case of a multiselect being identified as a leaf a '---LABEL' suffix is appended to the key and values of that key are returned
    *      form= { 'list': ['code1', 'code2'],
    *              'list---LABEL': ['label1', 'label2'] },
    *      fieldKey=list,
@@ -447,7 +447,7 @@ export class FormValueService {
       }
     }
 
-    // Clear out any MultiSelec labels.
+    // Clear out any MultiSelect labels.
     FormValueService.removeMultiSelectLabels(data);
   }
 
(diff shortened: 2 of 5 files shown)


src/shared/services/fields/fields.utils.ts
```diff
@@ -15,7 +15,8 @@ export class FieldsUtils {
 
   private static readonly currencyPipe: CurrencyPipe = new CurrencyPipe('en-GB');
   private static readonly datePipe: DatePipe = new DatePipe(new FormatTranslatorService());
-  public static readonly LABEL_SUFFIX = '-LABEL';
```
Daniel-Lam, line 18: @paul-graham do you think it's possible any Services might have tests that depend on label fields ending in `-LABEL`? Might be something we need to communicate to them, in case it breaks their tests - a bit like what happened with the change in some element IDs.

src/shared/services/form/form-value.service.ts
```diff
@@ -6,14 +6,12 @@ import { FieldTypeSanitiser } from './field-type-sanitiser';
 
 @Injectable()
 export class FormValueService {
-  public static readonly LABEL_SUFFIX = '-LABEL';
-
   /**
    * Gets value of a field based on fieldKey which is a dot separated reference to value and collection index.
    * There are two exeptions:
-   * 1) In case of a multiselect being identified as a leaf a '-LABEL' suffix is appended to the key and values og that key are returned
+   * 1) In case of a multiselect being identified as a leaf a '---LABEL' suffix is appended to the key and values og that key are returned
```
Daniel-Lam, line 12: @paul-graham - a nit-pick, I `og`led the comment and suspect it's meant to be `of`, instead 😄

src/shared/services/form/form-value.service.ts
```diff
@@ -422,6 +446,9 @@ export class FormValueService {
         }
       }
     }
+
+    // Clear out any MultiSelec labels.
```
Daniel-Lam, line 450: @paul-graham - another nit-pick... "Bo' MultiSelec**t**a" 😄

Daniel-Lam (changes requested):
@paul-graham - some very, very minor comments but otherwise all OK.

src/shared/services/form/form-value.service.ts
paul-graham: I changed that but never pushed it! 😂

paul-graham: Pushed now.

src/shared/services/form/form-value.service.ts
paul-graham: Changed. 🙂

Daniel-Lam (approved):

src/shared/services/fields/fields.utils.ts
paul-graham: This is held entirely internally so there shouldn't be an issue with this - there are no controls appearing anywhere on the screen with a `-LABEL` (or now `---LABEL`) suffix. It's simply a lookup for `MultiSelect` fields.

Daniel-Lam: Ah, OK - great. Thanks for explaining, @paul-graham.

adnanakgun (approved):

PR #856, opened 2021-07-14
Release 4.1.2 - ie fix for date components

leepavlouhmcts:
https://tools.hmcts.net/jira/browse/EUI-4279

### Change description ###
downgraded to "@angular-material-components/datetime-picker": "^2.0.4" and "@angular-material-components/moment-adapter": "^4.0.1" for ie 11 fix

```
[ ] Yes
[ ] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "4.1.0-rc.10",
+  "version": "4.1.0-rc.11",
   "engines": {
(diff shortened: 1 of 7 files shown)


Mo-Lala84 (approved):
approved

paul-graham (approved):
Looks like a perfectly reasonable downgrade of the angular material datetime-picker and moment-adapter. The latter is dropping down a major version, but it looks like this is internal only and with no external difference, hence not a breaking change for this library.

PR #857, opened 2021-07-15
EUI-2976: Strategic Activity Tracker

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-2976 (this is the epic)

### Change description ###
Introduced a new socket.io-based mechanism for tracking case activity. This sits alongside the existing polling mechanism and there is a feature flag to choose between the various modes (`'off'`, `'polling'`, and `'socket'`).

```
[ ] Yes
[x] No
```

### Notes ###
- There are still some unit tests to add in here and some refactoring to be done. This is mostly to get eyes on the PR.
- There is also a circular dependency warning coming up for `services/activity/activity-socket.service.ts` and `services/activity/index.ts` but I can't for the life of me see why. There's a prize for whoever figures out what's going on!
- I've checked `[x] No` above for the breaking change. That _should_ be correct but if you spot anything in there that looks like it's going to break MC, please let me know and I'll see if I can refactor it so it doesn't happen. If I can't, this will be a major release instead.
- Paired with https://github.com/hmcts/rpx-xui-webapp/pull/1172.


src/shared/components/activity/activity-banner/activity-banner.component.scss
```diff
@@ -1,40 +1,31 @@
...
     color: #FFFFFF;
     padding-left: 9px;
     position: relative;
     top: 50%;
     transform: translateY(-40%);
-}
-
-.bannerText {
+  }
+
```
adnanakgun, line 17: Did you run `yarn lint`? Extra white space here.

src/shared/components/search-result/search-result.component.html
```diff
@@ -31,7 +31,7 @@ <h2 class="heading-h2" id="search-result-heading__text" tabindex="-1">{{ caseSta
           </tr>
         </table>
       </th>
-      <th *ngIf="activityEnabled()" style="width: 110px;"></th>
+      <th *ngIf="activityEnabled" style="width: 110px;"></th>
```
adnanakgun, line 34: Maybe move inline styling to css?

adnanakgun (approved):

src/shared/components/activity/activity-banner/activity-banner.component.scss
paul-graham: No lint issues shown. I think it's because this is a `.scss` file and it only really cares about extra whitespace in `.ts` files...?

src/shared/components/search-result/search-result.component.html
paul-graham: Good spot. Inherited the code but I should have tidied it up while I was there. 👍

src/shared/components/activity/activity-banner/activity-banner.component.scss
adnanakgun: Possible.

src/shared/components/search-result/search-result.component.html
paul-graham: Done and pushed.

adnanakgun (dismissed):

adnanakgun (approved):

src/shared/services/activity/activity-socket.service.ts
```diff
@@ -0,0 +1,108 @@
...
+      .subscribe(mode => {
+        this.destroy();
+        if (ActivitySocketService.SOCKET_MODES.indexOf(mode) > -1) {
+          this.init();
+        }
+      });
+  }
+
+  public watchCases(caseIds: string[]): void {
+    const payload: WatchCases = { caseIds };
```
kuda1992, line 50: Surely this is better inline `this.socket.emit('watch', {caseIds})`

src/shared/services/activity/activity-socket.service.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  private getObservableOnSocketEvent<T>(event: string): Observable<T> {
+    return new Observable<T>(observer => {
+      this.socket.on(event, (payload: T) => {
+        observer.next(payload);
+      });
+    });
+  }
+
+  private setupUser(): object {
```
kuda1992, line 99: Can we add a `User` type instead of just leaving the return type as object

src/shared/services/activity/activity.polling.service.ts
```diff
@@ -105,17 +119,31 @@ export class ActivityPollingService {
...
+    this.pollConfig = {
+      interval: this.config.getActivityNexPollRequestMs(),
+      attempts: this.config.getActivityRetry(),
+      backgroundPolling: true
+    };
+    this.batchCollectionDelayMs = this.config.getActivityBatchCollectionDelayMs();
+    this.maxRequestsPerBatch = this.config.getActivityMaxRequestPerBatch();
+  }
+
+  private destroy(): void {
```
kuda1992, line 140: Empty method body. Do we need the method?

kuda1992 (changes requested):
Mostly good just left a few minor changes.

src/shared/components/case-viewer/case-event-trigger/case-event-trigger.component.ts
```diff
@@ -41,13 +51,24 @@ export class CaseEventTriggerComponent implements OnInit, OnDestroy {
...
-      this.ngZone.runOutsideAngular( () => {
-        this.activitySubscription = this.postEditActivity().subscribe((_resolved) => {
-          // console.log('Posted EDIT activity and result is: ' + JSON.stringify(_resolved));
-        });
+
+    this.activityService.modeSubject
+      .pipe(filter(mode => !!mode))
+      .pipe(distinctUntilChanged())
+      .subscribe(mode => {
+        if (ActivitySocketService.SOCKET_MODES.indexOf(mode) > -1) {
```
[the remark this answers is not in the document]
paul-graham: How would you propose doing that in this instance, @kuda1992? I toyed around with a few ways of doing this and this ended up being the cleanest version I could get to work... other than moving the nested stuff to somewhere else.

src/shared/services/activity/activity.polling.service.ts
paul-graham: Good spot. I realised that the `stopPolling()` method did what I needed _this_ method to do... so this is redundant and should be ditched.

src/shared/services/activity/activity-socket.service.ts
paul-graham: Agreed. I did this when I thought there'd be more required in the call and just never switched it to something simpler. Doing that now. The same for the other similar methods below.

src/shared/services/activity/activity-socket.service.ts
paul-graham: It'll be `UserInfo` as that's what it actually is so far as I can tell... but, yes, I'll make the change.

src/shared/components/activity/case-activity.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  selector: 'ccd-activity-icon',
+  inputs: ['description', 'imageLink']
+});
+
+const ActivityBannerComponent: any = MockComponent({
+  selector: 'ccd-activity-banner',
+  inputs: ['description', 'imageLink', 'bannerType']
+});
+
+xdescribe('CaseActivityComponent', () => {
```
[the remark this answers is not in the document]
paul-graham: These are now done and pushed.

src/shared/services/activity/activity-socket.service.ts
paul-graham: Done.

src/shared/services/activity/activity-socket.service.ts
paul-graham: Done.

src/shared/services/activity/activity.polling.service.ts
paul-graham: Done.

src/shared/services/activity/activity.service.ts
```diff
@@ -66,22 +75,22 @@ export class ActivityService {
     const options = this.getOptions();
     const url = `${this.activityUrl}/cases/${caseId.join(',')}/activity`;
     return this.http
-      .get(url, options, false)
+      .get(url, options, false, ActivityService.handleHttpError)
       .map(response => response);
   }
 
-  public postActivity(caseId: string, activityType: String): Observable<Activity[]> {
+  public postActivity(caseId: string, activity: String): Observable<Activity[]> {
```
kuda1992, line 82: `string` not `String` . I know this not you but we could tidy it up.

src/shared/services/activity/activity.service.ts
```diff
@@ -66,22 +75,22 @@ export class ActivityService {
...
     const options = this.getOptions();
     const url = `${this.activityUrl}/cases/${caseId}/activity`;
-    let body = { activity: activityType};
+    let body = { activity };
     return this.http
       .post(url, body, options, false)
       .map(response => response);
```
kuda1992, line 88: I don't understand why we need a `map` here? I have seen it many times on this project does it solve some kind of bug or something. If it does we need a comment.

src/shared/services/http/http-error.service.ts
```diff
@@ -48,10 +39,26 @@ export class HttpErrorService {
...
+  }
+
+  constructor(private readonly authService: AuthService) {}
+
+  public setError(error: HttpError) {
+    this.error = error;
+  }
+
+  public removeError(): HttpError {
+    let error = this.error;
```
kuda1992, line 52: Hmmm `let` . I think @Daniel-Lam would prefer `const`.

kuda1992 (changes requested):
Thanks @paul-graham for adding the tests and creating `Modes` enum. Left minor comments but looks mostly good.

src/shared/services/activity/activity.service.ts
paul-graham: Oh! I thought I'd sorted that one already! 😱

Might have been on a different branch but I'll make sure it's all good here too. Thank you.

src/shared/services/activity/activity.service.ts
paul-graham: An excellent point. This code was already in place and I didn't really assess it in a huge amount of detail as it was working fine... but you're right - the `map` is a no-op and so not needed at all. I'll change it.

src/shared/services/http/http-error.service.ts
paul-graham: Yep, agreed.

kuda1992 (approved):
Looks good. Thanks for making the changes @paul-graham 👍

adnanakgun (approved):

adnanakgun (approved):

kuda1992 (approved):

PR #869, opened 2021-07-23
EUI-4226: Activity tracker polls unauthorised users continuously

paul-graham:
https://tools.hmcts.net/jira/browse/EUI-4226

### Change description ###
Fixed an issue with the error message coming from the Activity Tracker coming back with a 403 but it getting stripped out in the `HttpError.from(...)` method.

```
[ ] Yes
[x] No
```

### Note ###
Linked to https://github.com/hmcts/rpx-xui-webapp/pull/1200.

### DIFF
--- a/src/shared/services/activity/activity.service.ts
+++ b/src/shared/services/activity/activity.service.ts
@@ -1,10 +1,11 @@
-import { Injectable } from '@angular/core';
-import { Activity } from '../../domain/activity';
-import { Observable } from 'rxjs';
-import { AbstractAppConfig } from '../../../app.config';
-import { HttpService, OptionsType } from '../../services/http';
 import { HttpHeaders } from '@angular/common/http';
-import { SessionStorageService } from '../session/session-storage.service';
+import { Injectable } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { AbstractAppConfig } from '../../../app.config';
+import { Activity } from '../../domain/activity';
+import { HttpService, OptionsType } from '../http';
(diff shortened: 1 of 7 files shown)


leepavlouhmcts (approved):

src/shared/services/activity/activity.service.ts
```diff
@@ -13,11 +14,17 @@ export class ActivityService {
   static get ACTIVITY_VIEW() { return 'view'; }
   static get ACTIVITY_EDIT() { return 'edit'; }
 
-  private userAuthorised;
+  private userAuthorised: boolean = undefined;
```
paulhowes-HMCTS, line 17: Ohh an undefined boolean. Not seen one of these in a while!

paulhowes-HMCTS (approved):
A bit of tidying up and an undefined boolean eh! Approved, despite the undefined boolean looking a little strange (and being the first one in ccd-case-ui-toolkit and MC!)

src/shared/services/activity/activity.service.ts
paul-graham: It was an untyped variable... so I gave it a type... but then the logic required that it be `undefined` before proceeding. I suppose I could have used some magic ~beans~ strings instead... 😂

paul-graham: Besides, it's usually considered a good thing to be first. Do I get a 🥇 now?

Mo-Lala84 (approved):
approved

leepavlouhmcts (approved):

PR #885, opened 2021-08-09
EUI-4329-Several events for Divorce Bulk Cases fail case data validation via the XUI when all is ok via CCD data validation

leepavlouhmcts:
https://tools.hmcts.net/jira/browse/EUI-4329

### Change description ###
added an additional formcontrol to the linkedCases formgroup to help identify when the abstractControl is of type linkedCases, then added an if statement further down the chain that correctly sets the value of 'CaseReference'

```
[ ] Yes
[x ] No
```


src/shared/components/palette/base-field/abstract-form-field.component.ts
```diff
@@ -65,10 +65,14 @@ export abstract class AbstractFormFieldComponent {
...
-      parent.setControl(this.caseField.id, control);
+      if (control['controls'] && control['controls']['Id'] && control['controls']['Id'].value === 'LinkedCase') {
+        parent.setControl('CaseReference', control['controls']['CaseReference']);
+      } else {
+        parent.setControl(this.caseField.id, control);
+      }
     } else {
       parent.addControl(this.caseField.id, control);
-    }
+    };
```
adnanakgun, line 75: Semi colon after a closing if clause? It is probably not breaking anything but let's not do that as it isn't standard practise.

adnanakgun (changes requested):
One minor change requested.

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
```diff
@@ -127,7 +127,7 @@ export class DatetimePickerComponent extends AbstractFormFieldComponent implemen
       this.hideTime = true;
       this.checkTime = false;
       this.dateTimeEntryFormat = this.formatTranslationService.removeTime(this.dateTimeEntryFormat);
-      this.momentFormat = 'YYYY-MM-DD';
+      this.momentFormat = 'YYYY-MM-DD'
```
kuda1992, line 130: Removed the semicolon why?

kuda1992 (changes requested):
I don't see any new tests updated for component `write-case-link-field.component.ts` to test the new logic.

src/shared/components/palette/base-field/abstract-form-field.component.ts
leepavlouhmcts: i will remove this now.

src/shared/components/palette/datetime-picker/datetime-picker.component.ts
leepavlouhmcts: that was a mistake i think with the merge. will put the semi colon back

leepavlouhmcts:
> I don't see any new tests updated for component `write-case-link-field.component.ts` to test the new logic.

added.

kuda1992 (dismissed):
Thanks for adding the tests @leepavlouhmcts . Approved 😊

adnanakgun (dismissed):

PR #891, opened 2021-08-11
EUI-4343-CCD case field validation indicator missing in WA2 release 2 part2 branch

kuda1992:
https://tools.hmcts.net/jira/browse/EUI-4343

### Change description ###
Added .touched property to the write-input  components to ensure the ```form-group-error``` class is appplied

```
[ ] Yes
[x ] No
```


src/shared/components/palette/case-link/write-case-link-field.html
```diff
@@ -1,8 +1,8 @@
-<div class="form-group" [ngClass]="{'form-group-error': !caseReferenceControl.valid && caseReferenceControl.dirty}">
+<div class="form-group" [ngClass]="{'form-group-error': !caseReferenceControl.valid && (caseReferenceControl.dirty || caseReferenceControl.touched)}">
```
rayliangatos, line 1: !caseReferenceControl.valid && (caseReferenceControl.dirty || caseReferenceControl.touched)}

this is doing the same job as:

caseReferenceControl.errors && (caseReferenceControl.dirty || caseReferenceControl.touched)

src/shared/components/palette/case-link/write-case-link-field.html
```diff
@@ -1,8 +1,8 @@
...
   <label [for]="id()">
     <span class="form-label" *ngIf="caseField.label">{{caseField | ccdFieldLabel}}</span>
   </label>
   <span class="form-hint" *ngIf="caseField.hint_text">{{caseField.hint_text}}</span>
-  <span class="error-message" *ngIf="caseReferenceControl.errors && caseReferenceControl.dirty">{{caseReferenceControl.errors | ccdFirstError:caseField.label}}</span>
+  <span class="error-message" *ngIf="caseReferenceControl.errors && (caseReferenceControl.dirty || caseReferenceControl.touched)">{{caseReferenceControl.errors | ccdFirstError:caseField.label}}</span>
```
rayliangatos, line 6:
```suggestion
extract the method to typescript:
public isControlInvalid(): boolean {
return caseReferenceControl.invalid && (caseReferenceControl.dirty || caseReferenceControl.touched);
}

Then replace it in the two places in html:
*ngIf="isControlInvalid"
and:
[ngClass]="{'form-group-error':isControlInvalid}"

This will avoid duplicated caculation in html, improve the html performance a bit.
Also it will be more concise in the code.
```

rayliangatos (changes requested):
left request changes comment

src/shared/components/palette/case-link/write-case-link-field.html
kuda1992: @rayliangatos I don't think I agree with you on improving performance yes it might be more concise. Using an angular method will cause angular to evaluate the method multiple times. Try still in a console.log in a method you can observe the issue.

rayliangatos: it's nothing related to whether you put the logic in the typescript or html, that is the ngIf directive's behaviour, it is asyn and "watches" for the variable's value. Even now you put all these in hmtl, it is still listening to the control's status all the time, it's just you couldn't console log this in html.

kuda1992: https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496

rayliangatos (approved):

RiteshHMCTS (approved):

PR #917, opened 2021-09-15
Do not merge: adding EventStartGuard, EventStartComponent with Statemanagment and Portal

udaydenduluri33:
https://tools.hmcts.net/jira/browse/EUI-4185
https://tools.hmcts.net/jira/browse/EUI-4187

### Change description ###
New  EventStartGuard and EventStartComponent

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/event-start/event-start-component.module.ts
+++ b/src/shared/components/event-start/event-start-component.module.ts
@@ -1,0 +1,27 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { ComponentPortalExample1Component, ComponentPortalExample2Component, EventStartComponent } from './event-start.component';
+import { PortalModule } from '@angular/cdk/portal';
+
+@NgModule({
+  imports: [
+    CommonModule,
+    PortalModule
+  ],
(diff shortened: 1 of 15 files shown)


src/shared/components/event-start/event-start-component.module.ts
```diff
@@ -0,0 +1,27 @@
...
+    ComponentPortalExample2Component
+  ],
+  entryComponents: [
+    EventStartComponent,
+    ComponentPortalExample1Component,
+    ComponentPortalExample2Component
+  ],
+  exports: [
+    EventStartComponent,
+    ComponentPortalExample1Component,
```
ronaldmansveld, line 23: Unless the components for the individual steps that are rendered within the portal are to be used outside of this module as well, there is no need to export them.

ronaldmansveld (dismissed):

src/shared/components/event-start/event-start-component.module.ts
udaydenduluri33: Point noted and removed it now

ronaldmansveld (dismissed):

PR #918, opened 2021-09-20
EUI-3473: Challenged Access Request screen for Case Access Management

Daniel-Lam:
https://tools.hmcts.net/jira/browse/EUI-3473

### Change description ###
Implementation of display-only User Story.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-viewer/viewer.routing.ts
+++ b/src/shared/components/case-viewer/viewer.routing.ts
@@ -5,6 +5,7 @@ import { CaseEventTriggerComponent } from './case-event-trigger/case-event-trigg
 import { editorRouting } from '../case-editor';
 import { CaseHistoryComponent } from '../case-history';
 import { FileUploadProgressGuard } from '../palette/document/file-upload-progress.guard';
+import { CaseChallengedAccessRequestComponent } from './case-challenged-access-request';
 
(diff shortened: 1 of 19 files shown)


rayliangatos (approved):
looks good, approved.

src/shared/components/case-viewer/viewer.routing.ts
```diff
@@ -23,5 +24,9 @@ export const viewerRouting: Routes = [
     {
       path: 'event/:eid/history',
       component: CaseHistoryComponent,
+    },
+    {
+      path: 'access-request',
```
adnanakgun, line 29: Do you think path should have the word `challenged` in it? Just a suggestion.

adnanakgun (dismissed):
Approved with a suggestion

src/shared/components/case-viewer/viewer.routing.ts
Daniel-Lam: @adnanakgun - could be an idea... although, I'm not sure if we're going to reuse this component for the "Specific Access" case - in which case, they _may_ want to use the same route... perhaps.

adnanakgun (approved):

rayliangatos (approved):

PR #966, opened 2021-12-03
Hotfix/e UI 4873 SSCS - Dynamic lists in a collection within a complex type

adnanakgun:

https://tools.hmcts.net/jira/browse/EUI-4873

### Change description ###

Reapplied dynamic list values for mid event callbacks using page wizard

```
[ ] Yes
[x] No
```


src/shared/services/fields/fields.utils.ts
```diff
@@ -242,6 +247,82 @@ export class FieldsUtils {
...
+   * down the tree. Server response returns data only in
+   * the `value` object of parent complex type
+   *
+   * EUI-2530 Dynamic Lists for Elements in a Complex Type
+   *
+   * @param jsonBody - { case_fields: [ CaseField, CaseField ] }
+   */
+   public static handleNestedDynamicLists(jsonBody: { case_fields: CaseField[] }): any {
+
+    if (jsonBody.case_fields) {
```
paulhowes-HMCTS, line 262: Is the if statement necessary? i.e. if there are none it just won't go into the foreach.

paulhowes-HMCTS (approved):
Just a minor tweak if you feel compelled to do it but no show stoppers by the look of it.

Mo-Lala84 (approved):
approved

src/shared/services/fields/fields.utils.ts
adnanakgun: I copy-pasted this code from CasesService to a more common location. Additionally, I sort of remember `foreach` causing an error when `case_fields` doesn't exist.

PR #1033, opened 2022-02-17
EUI-5225 organisation search bugfix

SivaprakasamSivaprakasam:
https://tools.hmcts.net/jira/browse/EUI-5225

### Change description ###
Organisation search optimization

```
[ ] Yes
[x] No
```


RiteshHMCTS (approved):

src/shared/components/palette/organisation/write-organisation-field.component.ts
```diff
@@ -125,12 +125,30 @@ export class WriteOrganisationFieldComponent extends AbstractFieldWriteComponent
...
+                          .find(item => item.organisationIdentifier === org.organisationIdentifier)
+      if (!matchingOrg && [...partMatchingResultSet, ...withSpaceMatchingResultSet].length < MAX_RESULT_COUNT
+        && this.searchWithSpace(org, lowerOrgSearchText)
+      ) {
+        withSpaceMatchingResultSet.push(org);
+      }
+    });
+    return [...partMatchingResultSet, ...withSpaceMatchingResultSet].map((organisation) =>
+      this.organisationConverter.toSimpleOrganisationModel(organisation)
+    );
```
rayliangatos, line 151: This seems like only for exactly search for the first time, if nothing is found then do the fuzzy search with space, if found something don't do the fuzzy search with space.
And two forEach here seems not necessary, probably there is potentially some performance issue occur.
As chat yesterday I think the easier way is first verify whether the search string is a post code pattern using Regex, if yes remove the space in the search string and just apply the exactly search here then it will be easier. The original searchWithSpace purpost is actually for searching the postcode. 
It will be good to retest this ticket: https://tools.hmcts.net/jira/browse/EUI-2087 for this change.
But search is a big topic, it depends whether user wants this searchWithSpace, if they don't like it we can provide search with exact word, that will get everything to be easier.

SivaprakasamSivaprakasam: @rayliangatos discussed a few options with Andy on the last design drop session and we have briefly agreed of going this way in a simpler way instead of redoing the whole stuff.

RiteshHMCTS (approved):

rayliangatos (approved):

RiteshHMCTS (dismissed):

src/shared/components/palette/organisation/write-organisation-field.component.ts
```diff
@@ -125,12 +125,31 @@ export class WriteOrganisationFieldComponent extends AbstractFieldWriteComponent
...
+  // 1. go through collection of org items one by one by doing the comparsion of search string using includes to all the address fields
+  // 2. split the search string into arrays and apply the each array item into the address fields
+  // 3. both step 1, 2 will go until max count result reaches, and finally combine both result sets into final collection
   public searchOrg(organisations: OrganisationVm[], lowerOrgSearchText: string): SimpleOrganisationModel[] {
-    return organisations.filter(organisation => {
-      return this.searchCriteria(organisation, lowerOrgSearchText) || this.searchWithSpace(organisation, lowerOrgSearchText);
-    })
-      .map(organisation => this.organisationConverter.toSimpleOrganisationModel(organisation))
-      .slice(0, WriteOrganisationFieldComponent.MAX_RESULT_COUNT);
+    let partMatchingResultSet = [], withSpaceMatchingResultSet = [];
```
phillip-whitaker-hmcts, line 133: these can be const as you don't redefine them, just push into them

src/shared/components/palette/organisation/write-organisation-field.component.ts
```diff
@@ -125,12 +125,31 @@ export class WriteOrganisationFieldComponent extends AbstractFieldWriteComponent
...
+        partMatchingResultSet.push(organisation);
+      }
+    });
+    organisations.forEach((org) => {
+      const matchingOrg = [...partMatchingResultSet, ...withSpaceMatchingResultSet]
+                          .find(item => item.organisationIdentifier === org.organisationIdentifier)
+      if (!matchingOrg && partMatchingResultSet.length === 0
+        && [...partMatchingResultSet, ...withSpaceMatchingResultSet].length < MAX_RESULT_COUNT
+        && this.searchWithSpace(org, lowerOrgSearchText)
+      ) {
```
phillip-whitaker-hmcts, line 146: There is a huge amount of stuff going on in the if statement. I would like to see this split up and made a bit more readable like and you repeat the concat of those two results, maybe change to this...

```
const resultSet =  [...partMatchingResultSet, ...withSpaceMatchingResultSet];
const hasMatchingOrganisation = resultSets .find(item => item.organisationIdentifier === org.organisationIdentifier);
const searchHasSpace = this.searchWithSpace(org, lowerOrgSearchText);
const resultSetLengthBelowMax = resultSet .length < MAX_RESULT_COUNT;


if (!hasMatchingOrganisation && .. )
```

src/shared/components/palette/organisation/write-organisation-field.component.ts
```diff
@@ -125,12 +125,31 @@ export class WriteOrganisationFieldComponent extends AbstractFieldWriteComponent
...
+        withSpaceMatchingResultSet.push(org);
+      }
+    });
+    return [...partMatchingResultSet, ...withSpaceMatchingResultSet].map((organisation) =>
+      this.organisationConverter.toSimpleOrganisationModel(organisation)
+    );
```
phillip-whitaker-hmcts, line 152: with the above refactor, this could be simplified to

`return resultSet.map(this.organisationConverter.toSimpleOrganisationModel)`

phillip-whitaker-hmcts (changes requested):
I think the code could be cleared up a bit.

src/shared/components/palette/organisation/write-organisation-field.component.ts
RiteshHMCTS: @phillip-whitaker-hmcts address this rectification

src/shared/components/palette/organisation/write-organisation-field.component.ts
RiteshHMCTS: cannot change to resultSet as withSpaceMatchingResultSet is getting updated after

phillip-whitaker-hmcts (dismissed):
approved

phillip-whitaker-hmcts (approved):

johnbenjamin-hmcts (approved):
Approved

PR #1045, opened 2022-02-28
EUI-4897: Case flags tab - banner message display

Daniel-Lam:
[EUI-4897](https://tools.hmcts.net/jira/browse/EUI-4897)

### Change description ###
Fix case flags tab navigation from message banner.

```
[ ] Yes
[x] No
```


johnbenjamin-hmcts (approved):
Approved

src/shared/components/case-viewer/case-full-access-view/case-full-access-view.component.ts
```diff
@@ -250,12 +255,26 @@ export class CaseFullAccessViewComponent implements OnInit, OnDestroy, AfterView
...
     }
   }
 
+  public getCaseFlagsTabName(): string {
+    // Determine which tab contains the FlagLauncher CaseField type, from the CaseView object in the snapshot data,
+    // and get its label
+    if (!this.caseFlagsTabName && this.route.snapshot.data.case && this.route.snapshot.data.case.tabs) {
+      this.caseFlagsTabName = (this.route.snapshot.data.case.tabs as CaseTab[])
+      .filter(tab => tab.fields && tab.fields
+        .some(caseField => caseField.field_type.id === 'FlagLauncher' && caseField.field_type.type === 'FlagLauncher'))[0].label;
```
rayliangatos, line 272: Is that field_type.id === 'FlagLauncher' enough to get the filed_type which is 'FlagLauncher'? Do we need to judge that again from the caseField.field_type.type which seems duplicated, does it have this case when that id === 'FlagLauncher' but type is different?

Daniel-Lam: hi @rayliangatos - I have checked a bit further and it does seem there are cases where `field_type.id` and `field_type.type` are not the same. Examples include collections and Complex types. I will change the check to look at `field_type.type` only.

rayliangatos (approved):
thanks for addressing the comment, looks good to me.

RiteshHMCTS (approved):

src/shared/components/palette/case-flag/read-case-flag-field.component.html
```diff
@@ -4,4 +4,4 @@ <h3 class="govuk-heading-m">Party level flags</h3>
   <ccd-case-flag-table [flagData]="flagData"></ccd-case-flag-table>
 </div>
 <h3 class="govuk-heading-m">Case level flags</h3>
-<ccd-case-flag-table [flagData]="caseLevelCaseFlagData"></ccd-case-flag-table>
+<!-- <ccd-case-flag-table [flagData]="caseLevelCaseFlagData"></ccd-case-flag-table> -->
```
RiteshHMCTS, line 7: @Daniel-Lam can you remove this commented code

johnbenjamin-hmcts (approved):
Approved

src/shared/components/palette/case-flag/read-case-flag-field.component.html
Daniel-Lam: @RiteshHMCTS - this will be re-enabled at some point in the future, when we have case-level Case Flags data, hence left as a placeholder.

RiteshHMCTS: then its fine @Daniel-Lam

PR #1062, opened 2022-03-23
Withhold certain errors as per ticket

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EUI-5528

### Change description ###
Withholding errors for certain cases so they do not go to default url when the case service is not yet supported by WA

```
[ ] Yes
[ x ] No
```


src/shared/components/case-editor/services/work-allocation.service.ts
```diff
@@ -178,7 +179,15 @@ export class WorkAllocationService {
    * @param caseId The ID of the case to find tasks for.
    */
   public getTasksByCaseIdAndEventId(eventId: string, caseId: string, caseType: string, jurisdiction: string): Observable<TaskPayload> {
-    return this.http.get(`${this.appConfig.getWorkAllocationApiUrl()}/case/tasks/${caseId}/event/${eventId}/caseType/${caseType}/jurisdiction/${jurisdiction}`);
+    return this.http.get(`${this.appConfig.getWorkAllocationApiUrl()}/case/tasks/${caseId}/event/${eventId}/caseType/${caseType}/jurisdiction/${jurisdiction}`).pipe(catchError(error => {
+      if (error instanceof HttpErrorResponse) {
+        if (error.status === 401 || error.status === 403 || error.status === 404) {
```
udaydenduluri33, line 184: can check for error before error.status

udaydenduluri33 (approved):

src/shared/components/case-editor/services/work-allocation.service.ts
scott-turner-hmcts: @connorpgpmcelroy Also if there is no code to be executed in the if statement should it be re-written, something like this?

```
const { status } = error;
if (status !== 401 && status !== 403 && status !== 404) {
  return throwError(error);
}
```

scott-turner-hmcts (approved):

PR #1067, opened 2022-03-29
EUI-5495/EUI-4845: Create Case Flag - Reference Data integration

Daniel-Lam:
[EUI-5495](https://tools.hmcts.net/jira/browse/EUI-5495)
[EUI-4845](https://tools.hmcts.net/jira/browse/EUI-4845)

### Change description ###
Integrate "select flag type" step of case flag creation with Reference Data Common API, to retrieve list of flag types.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/palette/case-flag/components/select-flag-type/select-flag-type.component.ts
+++ b/src/shared/components/palette/case-flag/components/select-flag-type/select-flag-type.component.ts
@@ -1,15 +1,19 @@
-import { Component, EventEmitter, Input, OnInit, Output } from '@angular/core';
+import { Component, EventEmitter, Input, OnDestroy, OnInit, Output } from '@angular/core';
 import { FormControl, FormGroup } from '@angular/forms';
-import { CaseFlagFieldState, CaseFlagWizardStepTitle, SelectFlagTypeErrorMessage } from '../../enums';
+import { Subscription } from 'rxjs';
 import { ErrorMessage } from '../../../../../domain';
-import { CaseFlagState, FlagType } from '../../domain';
+import { FlagType } from '../../../../../domain/case-flag';
+import { CaseFlagRefdataService } from '../../../../../services';
+import { RefdataCaseFlagType } from '../../../../../services/case-flag/refdata-case-flag-type.enum';
+import { CaseFlagState } from '../../domain';
+import { CaseFlagFieldState, CaseFlagWizardStepTitle, SelectFlagTypeErrorMessage } from '../../enums';
 
 @Component({
   selector: 'ccd-select-flag-type',
   templateUrl: './select-flag-type.component.html',
   styleUrls: ['./select-flag-type.component.scss']
 })
-export class SelectFlagTypeComponent implements OnInit {
+export class SelectFlagTypeComponent implements OnInit, OnDestroy {
 
   @Input()
   public formGroup: FormGroup;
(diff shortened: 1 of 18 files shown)


johnbenjamin-hmcts (changes requested):
@Daniel-Lam I think you have to create spec file for CaseFlagRefdataService.

Daniel-Lam:
> @Daniel-Lam I think you have to create spec file for CaseFlagRefdataService.

Hi @johnbenjamin-hmcts - you are, of course, correct... I ran out of time and was hoping to add it later 😄.

Kartheek-S (approved):

src/shared/services/case-flag/case-flag-refdata.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (url) {
+      url += '/orgServices';
+      if (serviceNames) {
+        url += `?ccdServiceNames=${serviceNames}`;
+      }
+
+      return this.http.get(url, {observe: 'body'});
+    }
+
+    return null;
```
rayliangatos, line 77: does this return null match Observable<HmctsServiceDetail[]>? or it could be return of([]);

src/shared/services/case-flag/case-flag-refdata.service.ts
```diff
@@ -0,0 +1,79 @@
...
+              // Note: Reference Data Common API appears to respond with a 404 error rather than send an empty response,
+              // so this may be redundant
+              return throwError(new Error('No flag types could be retrieved'));
+            }
+            return body.flags[0].FlagDetails;
+          })
+        );
+    }
+
+    return null;
```
rayliangatos, line 55: does this return null match Observable<FlagType[]>?

src/shared/components/palette/case-flag/components/select-flag-type/select-flag-type.component.ts
```diff
@@ -21,37 +25,70 @@ export class SelectFlagTypeComponent implements OnInit {
...
-    this.formGroup.addControl('flagType', new FormControl(''));
-    this.formGroup.addControl('other', new FormControl(''));
-    this.formGroup.addControl('otherFlagTypeDescription', new FormControl(''));
+    this.flagTypes = [];
+    this.formGroup.addControl(this.flagTypeControlName, new FormControl(''));
+    this.formGroup.addControl(this.descriptionControlName, new FormControl(''));
+
+    // HMCTS service code should come from Refdata API lookup via getHmctsServiceCode() but this is not working in AAT;
+    // use test service code 'AAA1' for now
+    this.flagRefdata$ = this.caseFlagRefdataService.getCaseFlagsRefdata('AAA1', RefdataCaseFlagType.PARTY).subscribe({
```
rayliangatos, line 55: What does this 'AAA1' hard code mean?

Daniel-Lam: @rayliangatos - 'AAA1' is a test service code to be used for looking up Case Flags refdata. It will be replaced in a subsequent PR.

rayliangatos: @Daniel-Lam okay, as I'm seeting this  'AAA1' is added to the actual logical ts file not the spec.ts test file, which confused me and I think it should be avoided, And it would be better to passed from somewhere correctly.

rayliangatos (approved):

PR #1078, opened 2022-04-14
EUI-4284 request created

tarikdogan-hmcts:

https://tools.hmcts.net/jira/browse/EUI-4284

### Change description ###

Node layer- technical ticket (request specific access )

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/domain/case-view/role-request.model.ts
+++ b/src/shared/domain/case-view/role-request.model.ts
@@ -20,7 +20,10 @@ export interface RequestedRole {
 }
 
 export interface RoleRequest {
-  assignerId: string
+  assignerId: any,
+  process: any,
+  replaceExisting: any,
+  reference:any
 }
 
 export interface RequestedRoleNote {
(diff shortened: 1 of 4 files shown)


src/shared/domain/case-view/role-request.model.ts
```diff
@@ -20,7 +20,10 @@ export interface RequestedRole {
 }
 
 export interface RoleRequest {
-  assignerId: string
+  assignerId: any,
```
connorpgpmcelroy, line 23: Can we use string for these or will these be objects? - prefer not to use any

connorpgpmcelroy (changes requested):
Add types for RoleRequest?

connorpgpmcelroy (dismissed):

RiteshHMCTS (approved):

connorpgpmcelroy (approved):
Approved on basis it builds correctly

waynepHMCTS (approved):

PR #1098, opened 2022-05-16
challenged access role integration changes

udaydenduluri33:

https://tools.hmcts.net/jira/browse/EUI-5844

### Change description ###

Challenged Access 

```
[ ] Yes
[ X ] No
```


src/shared/components/case-editor/case-access-utils/index.ts
```diff
@@ -63,7 +63,7 @@ export class CaseAccessUtils {
                 roleName = `${accessType}-access-${CaseAccessUtils.ADMIN_ROLE_NAME}`;
                 break;
             default:
-                roleName = `${accessType}-access-${CaseAccessUtils.LEGAL_OPERATIONS_ROLE_NAME}`;
+                roleName = `${accessType}-access-legal-ops`;
```
connorpgpmcelroy, line 66: Can't we change the variable name here that was being used?

src/shared/components/case-editor/case-access-utils/index.ts
```diff
@@ -85,15 +85,15 @@ export class CaseAccessUtils {
         const payload: RoleRequestPayload = {
             roleRequest: {
                 assignerId: assignerId,
-                process: 'specific-access',
-                reference: assignerId,
-                replaceExisting: false
+                process: 'challenged-access',
```
connorpgpmcelroy, line 88: Does this mean specific access requests will no longer work?

connorpgpmcelroy (changes requested):
Variable should be used, also looks like this may replace some specific access functionality

src/shared/components/case-editor/case-access-utils/index.ts
udaydenduluri33: Very good point. I did not realise that same function is being used by Specific access. I will change it now

src/shared/components/case-editor/case-access-utils/index.ts
udaydenduluri33: Done

src/shared/components/case-editor/case-access-utils/index.ts
udaydenduluri33: Done

connorpgpmcelroy (dismissed):
Looks good to me

tarikdogan-hmcts (dismissed):

tarikdogan-hmcts (approved):

PR #1136, opened 2022-06-15
EUI-5342 Update case flag

johnbenjamin-hmcts:
https://tools.hmcts.net/jira/browse/EUI-5342

### Change description ###
Manage case flags - Update flag page - Functionality and navigation

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/palette/case-flag/write-case-flag-field.component.ts
+++ b/src/shared/components/palette/case-flag/write-case-flag-field.component.ts
@@ -77,18 +77,18 @@ export class WriteCaseFlagFieldComponent extends AbstractFieldWriteComponent imp
                 partyName: caseField.value.partyName,
                 roleOnCase: caseField.value.roleOnCase,
                 details: caseField.value.details
-                  ? ((caseField.value.details) as any[]).map(detail => {
+                  ? ((caseField.value.details) as any[]).map((detail) => {
                     return Object.assign({}, ...Object.keys(detail.value).map(k => {
                       switch (k) {
                         // These two fields are date-time fields
                         case 'dateTimeModified':
                         case 'dateTimeCreated':
-                          return {[k]: new Date(detail.value[k])};
+                          return {[k]: new Date(detail.value[k]), 'id': detail.id};
                         // This field is a "yes/no" field
                         case 'hearingRelevant':
-                          return detail.value[k].toUpperCase() === 'YES' ? {[k]: true} : {[k]: false};
+                          return detail.value[k].toUpperCase() === 'YES' ? {[k]: true, 'id': detail.id} : {[k]: false, 'id': detail.id};
                         default:
-                          return {[k]: detail.value[k]};
+                          return {[k]: detail.value[k], 'id': detail.id};
                       }
                     }));
                   }) as FlagDetail[]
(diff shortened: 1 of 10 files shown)


tarikdogan-hmcts (approved):

src/shared/components/palette/case-flag/write-case-flag-field.component.ts
```diff
@@ -171,20 +180,46 @@ export class WriteCaseFlagFieldComponent extends AbstractFieldWriteComponent imp
...
-          flagsCaseFieldValue.details.push({value: this.populateNewFlagDetailInstance()});
+        switch (this.fieldState) {
+          case CaseFlagFieldState.FLAG_COMMENTS:
+            this.addFlagToCollection();
+            break;
+          case CaseFlagFieldState.FLAG_UPDATE:
+            this.updateFlagInCollection();
+            break;
+          default:
+            console.log('Invalid CaseFlagFieldState:', this.fieldState);
```
Daniel-Lam, line 191: @johnbenjamin-hmcts - I think we can remove the `console.log` statement because there shouldn't be any other states apart from these two

src/shared/components/palette/case-flag/write-case-flag-field.component.ts
```diff
@@ -171,20 +180,46 @@ export class WriteCaseFlagFieldComponent extends AbstractFieldWriteComponent imp
...
     }
   }
 
+  public addFlagToCollection(): void {
+    const flagsCaseFieldValue = this.caseFlagParentFormGroup['caseField'].value;
+    if (flagsCaseFieldValue) {
+      flagsCaseFieldValue.details.push({value: this.populateNewFlagDetailInstance()});
+      // There is no error, update form group value and validity
+      this.formGroup.updateValueAndValidity();
+    } else {
```
Daniel-Lam, line 204: @johnbenjamin-hmcts - not really sure how we should handle the `else` case. It shouldn't be the case that `flagsCaseFieldValue` is falsy, yet if it is then is there anything the user can do? Maybe we should fail silently for now...

src/shared/components/palette/case-flag/write-case-flag-field.component.ts
```diff
@@ -171,20 +180,46 @@ export class WriteCaseFlagFieldComponent extends AbstractFieldWriteComponent imp
...
+  public updateFlagInCollection(): void {
+    const flagsCaseFieldValue = this.caseFlagParentFormGroup['caseField'].value;
+    const flagDetailToUpdate = flagsCaseFieldValue.details.find(detail => detail.id === this.selectedFlagDetail.id);
+    if (flagDetailToUpdate) {
+      flagDetailToUpdate.value.flagComment = this.caseFlagParentFormGroup.value.flagComments
+        ? this.caseFlagParentFormGroup.value.flagComments
+        : null;
+      // There is no error, update form group value and validity
+      this.formGroup.updateValueAndValidity();
+    } else {
```
Daniel-Lam, line 218: @johnbenjamin-hmcts - not really sure how we should handle the `else` case. It shouldn't be the case that `flagDetailToUpdate` is falsy, yet if it is then is there anything the user can do? Maybe we should fail silently for now...

Daniel-Lam (changes requested):
@johnbenjamin-hmcts - just a few comments to look at

src/shared/components/palette/case-flag/write-case-flag-field.component.ts
johnbenjamin-hmcts: Thank you @Daniel-Lam I have now updated the code.

src/shared/components/palette/case-flag/write-case-flag-field.component.ts
johnbenjamin-hmcts: @Daniel-Lam I have now updated the code to delete the else part.

src/shared/components/palette/case-flag/write-case-flag-field.component.ts
johnbenjamin-hmcts: @Daniel-Lam I have now updated the code to delete the else part.

Daniel-Lam (approved):

PR #1178, opened 2022-08-23
bug/EUI-6599 case error information

Kartheek-S:
https://tools.hmcts.net/jira/browse/EUI-6599

### Change description ###
Error handing for APIs

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/src/shared/components/palette/case-link/components/read-linked-cases.component.html
+++ b/src/shared/components/palette/case-link/components/read-linked-cases.component.html
@@ -2,13 +2,31 @@
     <div class="govuk-!-margin-bottom-2">
         <h1 class="heading-h1">Linked cases</h1>
         <label class="table-sub-heading">This case is linked to</label>
-        <ccd-linked-cases-to-table [caseField]="caseField" (notifyAPIFailure)="getFailureNotification($event)"
+        <ccd-linked-cases-to-table [caseField]="caseField" (notifyAPIFailure)="getFailureLinkedToNotification($event)"
             reloadPage="reload"></ccd-linked-cases-to-table>
+        <ng-container *ngIf="serverLinkedToError && !serverError">
+          <div class="govuk-main-wrapper">
+              <p class="govuk-body">{{serverLinkedToError.message}}</p>
+              <span class="govuk-body align-left">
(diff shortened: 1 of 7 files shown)


src/shared/components/palette/case-link/components/read-linked-cases.component.html
```diff
@@ -2,13 +2,31 @@
     <div class="govuk-!-margin-bottom-2">
         <h1 class="heading-h1">Linked cases</h1>
         <label class="table-sub-heading">This case is linked to</label>
-        <ccd-linked-cases-to-table [caseField]="caseField" (notifyAPIFailure)="getFailureNotification($event)"
+        <ccd-linked-cases-to-table [caseField]="caseField" (notifyAPIFailure)="getFailureLinkedToNotification($event)"
```
SivaprakasamSivaprakasam, line 5: It's not just to separate the error handling for from and to table rendering and it's more of tracking how many apis failed and at the same time we should able to see if all the api's failed or some of them, don't think its happening now ?

SivaprakasamSivaprakasam (changes requested):
please check @Kartheek-S the comments as the implementation may need a change.

src/shared/components/palette/case-link/components/read-linked-cases.component.ts
```diff
@@ -18,51 +18,71 @@ export class ReadLinkedCasesComponent implements OnInit, AfterViewInit {
...
+  public getFailureLinkedFromNotification(evt) {
+    const errorMessage = 'Some case information is not available at the moment';
+    this.serverLinkedFromError = {
+      id: 'backendError', message: errorMessage
+    };
+    this.getFailureNotification(evt);
+  }
+
+  public getFailureNotification(evt) {
+    if (this.serverLinkedToError && this.serverLinkedFromError) {
```
rayliangatos, line 81: Is this or(||) rather than && when seeing either these two messages it is the backend error?

Kartheek-S: Updated BE error handling

rayliangatos (approved):

SivaprakasamSivaprakasam (approved):

PR #1186, opened 2022-08-25
EUI-6585

maliumer:
https://tools.hmcts.net/jira/browse/EUI-6585

### Change description ###
ccd-case-ui-toolkit Welsh translation list of templates

```
[ ] Yes
[X] No
```


projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/collection/read-collection-field.html
```diff
@@ -7,7 +7,7 @@
         <ccd-field-read
           [caseField]="{
             id: caseField.label,
-            label: caseField.label,
+            label: caseField.label | rpxTranslate,
```
rayliangatos, line 10: shouldn't this has been translated in the down level ccd-field-read component? What happened if you had this translated here and then you translate again in ccd-field-read? We should avoid translating in the customized component property but it should be in the html template which can't be broken down anymore.

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/collection/read-collection-field.html
```diff
@@ -25,7 +25,7 @@
         <ccd-field-read
           [caseField]="{
             id: i,
-            label: caseField.label + ' ' + (i + 1),
+            label: caseField.label + ' ' + (i + 1) | rpxTranslate,
```
rayliangatos, line 28: shouldn't this has been translated in the down level ccd-field-read component?

MateiHMCTS:
Not needed anymore

PR #1191, opened 2022-09-07
Feature/e UI 6714

SivaprakasamSivaprakasam:
https://tools.hmcts.net/jira/browse/EUI-6714

### Change description ###

https://tools.hmcts.net/jira/browse/EUI-6714

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/shared/components/palette/case-link/components/before-you-start/before-you-start.component.ts
+++ b/src/shared/components/palette/case-link/components/before-you-start/before-you-start.component.ts
@@ -10,7 +10,6 @@ import { LinkedCasesService } from '../../services/linked-cases.service';
   templateUrl: './before-you-start.component.html'
 })
 export class BeforeYouStartComponent {
-  public hasAnyAPIFailure = false;
 
   @Output()
   public linkedCasesStateEmitter: EventEmitter<LinkedCasesState> = new EventEmitter<LinkedCasesState>();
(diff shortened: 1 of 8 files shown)


maliumer (approved):

src/shared/components/palette/case-link/components/read-linked-cases.component.ts
```diff
@@ -68,9 +68,8 @@ export class ReadLinkedCasesComponent implements OnInit, AfterViewInit {
   }
 
   public reloadCurrentRoute() {
-    let currentUrl = this.router.url;
-    this.router.navigateByUrl('/', { skipLocationChange: true }).then(() => {
-      this.router.navigate([currentUrl]);
+    this.router.navigate(['cases', 'case-details', this.linkedCasesService.caseDetails.case_id]).then(() => {
+      window.location.hash = 'Linked cases';
```
rayliangatos, line 72: 1. suppose it shouldn't use blank string in the hash? 
2. and does it have that hash key in the html template?

src/shared/components/palette/case-link/components/before-you-start/before-you-start.component.ts
```diff
@@ -35,4 +34,10 @@ export class BeforeYouStartComponent {
       navigateToNextPage: true
     });
   }
+
+  public onBack(): void {
+    this.router.navigate(['cases', 'case-details', this.linkedCasesService.caseId]).then(() => {
+      window.location.hash = 'Linked cases';
```
rayliangatos, line 40: 1. suppose it shouldn't use blank string in the hash? 
2. and does it have that hash key in the html template?

SivaprakasamSivaprakasam: As far as I know, there is no hash key for the linked  cases and we usually redirect to the landing page tab.

src/shared/components/palette/case-link/components/read-linked-cases.component.ts
SivaprakasamSivaprakasam: As far as I know, there is no hash key for the linked  cases and we usually redirect to the landing page tab.

rayliangatos: so does that mean this window.location.hash never work can we remove it?

src/shared/components/palette/case-link/components/before-you-start/before-you-start.component.ts
rayliangatos: so does that mean this window.location.hash never work can we remove it?

rayliangatos (approved):

PR #1193, opened 2022-09-08
bug/EUI-6513

maliumer:
https://tools.hmcts.net/jira/browse/EUI-6513

### Change description ###
DisplayContext = READONLY doesn't work as expected

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/palette/base-field/field-write.component.ts
+++ b/src/shared/components/palette/base-field/field-write.component.ts
@@ -35,7 +35,7 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements 
   }
 
   ngOnInit(): void {
-    const writeMode = !(this.caseField.display_context === 'READONLY')
+    const writeMode = this.caseField.display_context !== 'READONLY';
     const componentClass = this.paletteService.getFieldComponentClass(this.caseField, writeMode);
 
     const injector = Injector.create([], this.fieldContainer.parentInjector);
--- a/src/shared/components/palette/base-field/field-write.component.ts
+++ b/src/shared/components/palette/base-field/field-write.component.ts
@@ -35,7 +35,8 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements 
   }
 
   ngOnInit(): void {
-    let componentClass = this.paletteService.getFieldComponentClass(this.caseField, true);
+    let writeMode = !(this.caseField.display_context === 'READONLY')
+    let componentClass = this.paletteService.getFieldComponentClass(this.caseField, writeMode);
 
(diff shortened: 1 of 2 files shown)


RiteshHMCTS (approved):

src/shared/components/palette/base-field/field-write.component.ts
```diff
@@ -35,7 +35,8 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements
   }
 
   ngOnInit(): void {
-    let componentClass = this.paletteService.getFieldComponentClass(this.caseField, true);
+    let writeMode = !(this.caseField.display_context === 'READONLY')
+    let componentClass = this.paletteService.getFieldComponentClass(this.caseField, writeMode);
```
Daniel-Lam, line 39: @maliumer - please can you add a basic unit test to `field-write.component.spec.ts` to check that `paletteService.getFieldComponentClass()` is being called with the correct parameters, based on the value of `writeMode`? In fact, you could probably do away with the separate `writeMode` variable - since it's a boolean - and just use the expression as the second argument to `getFieldComponentClass()`

Daniel-Lam (changes requested):
@maliumer - I've requested a small change plus the addition of a unit test

Daniel-Lam (changes requested):
@maliumer - just one more change needed

Daniel-Lam (changes requested):
@maliumer - suggestion for small change to unit test

Daniel-Lam (dismissed):

src/shared/components/palette/base-field/field-write.component.ts
```diff
@@ -35,10 +35,11 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements
   }
 
   ngOnInit(): void {
-    let componentClass = this.paletteService.getFieldComponentClass(this.caseField, true);
+    const writeMode = !(this.caseField.display_context === 'READONLY')
```
phillip-whitaker-hmcts, line 38: This seems a bit odd, why not just 

`const writeMode = this.caseField.display_context !== 'READONLY'`

Also, missing a semicolon

phillip-whitaker-hmcts (changes requested):
One small comment

src/shared/components/palette/base-field/field-write.component.ts
Daniel-Lam: @phillip-whitaker-hmcts - yeah... I did raise the same observation earlier, that perhaps `writeMode` is redundant

Daniel-Lam (approved):
@maliumer - approved

PR #1228, opened 2022-10-24
bug/EUI-6975 case exclusion

Kartheek-S:
https://tools.hmcts.net/jira/browse/EUI-6975

### Change description ###
User Excluded from case accessing same case via 16 digits reference within same session

```
[ ] Yes
[ x ] No
```


udaydenduluri33 (changes requested):
Minor change - please introduce a function removeCachedCase in CaseNotifier instead if marking it to null

src/shared/components/case-editor/services/case.notifier.ts
```diff
@@ -13,6 +13,10 @@ export class CaseNotifier {
 
     constructor(private casesService: CasesService) {}
 
+    public removeCahedCase() {
```
MateiHMCTS, line 16: typo

udaydenduluri33 (approved):

DavidJayakumar (approved):

PR #1234, opened 2022-10-28
Set basic fields

DavidJayakumar:

https://tools.hmcts.net/jira/browse/EUI-7006

### Change description ###

Set basic fields

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/shared/components/case-editor/services/case.notifier.ts
+++ b/src/shared/components/case-editor/services/case.notifier.ts
@@ -7,11 +7,11 @@ import { CasesService } from './cases.service';
 
 @Injectable()
 export class CaseNotifier {
+    public static readonly CASE_NAME = 'caseNameHmctsInternal';
+    public static readonly CASE_LOCATION = 'caseManagementLocation'
     private caseViewSource: BehaviorSubject<CaseView> = new BehaviorSubject<CaseView>(new CaseView());
     caseView = this.caseViewSource.asObservable();
     public cachedCaseView: CaseView;
-    public static readonly CASE_NAME = 'caseNameHmctsInternal';
-    public static readonly CASE_LOCATION = 'caseManagementLocation'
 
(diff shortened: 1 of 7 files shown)


Kartheek-S (approved):

udaydenduluri33 (approved):

src/shared/components/case-editor/services/case.notifier.ts
```diff
@@ -27,9 +29,22 @@ export class CaseNotifier {
...
             return this.cachedCaseView;
           }),
         );
     }
+    public setBasicFields(tabs: CaseTab[]): void {
+      tabs.forEach((t) => {
+        const caseName = t.fields.find(f => f.id === CaseNotifier.CASE_NAME);
+        const caseLocation = t.fields.find(f => f.id === CaseNotifier.CASE_LOCATION);
+        if (caseName || caseLocation) {
+          this.cachedCaseView.basicFields = {
```
andywilkinshmcts, line 43: I don't think this will work. If the fields are separate tabs, the second tab will clear out the field on the first tab. 
Also it's slightly inefficient, in that it inspects all tabs. I would do something like this:
tabs.some(t => {
        const caseName = t.fields.find(f => f.id === CaseNotifier.CASE_NAME);
        const caseLocation = t.fields.find(f => f.id === CaseNotifier.CASE_LOCATION);
        if (caseName && caseLocation.value) {
          this.cachedCaseView.basicFields.caseNameHmctsInternal = caseName.value;
        }
        if (caseLocation && caseLocation.value) {
          this.cachedCaseView.basicFields.caseManagementLocation = caseLocation.value;
        }
        return this.cachedCaseView.baseFields.caseName && this.cachedCaseView.basicFields.caseManagementLocation;

       }

andywilkinshmcts (commented):
I don't think this will work in all scenarios

PR #1305, opened 2023-01-06
EUI-7344 tab url fix

tarikdogan-hmcts:

https://tools.hmcts.net/jira/browse/EUI-7344

### Change description ###

Fix for when URL changes, tab not changes in some cases (link)

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/shared/components/case-viewer/case-full-access-view/case-full-access-view.component.ts
+++ b/src/shared/components/case-viewer/case-full-access-view/case-full-access-view.component.ts
@@ -119,7 +119,6 @@ export class CaseFullAccessViewComponent implements OnInit, OnDestroy, OnChanges
       });
     }
     this.router.events.subscribe((val) => {
-      console.log(val);
       const url = val && (val as any).url;
       if (url) {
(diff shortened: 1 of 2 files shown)


GaryScottHMCTS (approved):

src/shared/components/case-viewer/case-full-access-view/case-full-access-view.component.ts
```diff
@@ -118,7 +118,17 @@ export class CaseFullAccessViewComponent implements OnInit, OnDestroy, OnChanges
         this.activitySubscription = this.postViewActivity().subscribe();
       });
     }
-
```
phillip-whitaker-hmcts, line 121: My only comment is that I think this should be split out into a method, there is already a lot going on in this ngOnit().. just create a method - checkRouteAndSetCaseViewTab() and call it from in here.

phillip-whitaker-hmcts (approved):

GaryScottHMCTS (dismissed):

PR #1401, opened 2023-03-29
EXUI-110 Infinite Tab Loop Fix

deanbullock-hmcts:
### JIRA link  ###

https://tools.hmcts.net/jira/browse/EXUI-110

### Change description ###

Use selectedIndexChange instead of selectedTabChange on MatTabGroup fixes bug


Mo-Lala84 (dismissed):
approved

PR #1453, opened 2023-05-16
Feature/EUI 8139

DavidJayakumar:

https://tools.hmcts.net/jira/browse/EUI-8139

### Change description ###

Add sorting for QM table

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/components/query-list/query-list.component.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/components/query-list/query-list.component.ts
@@ -1,5 +1,6 @@
 import { Component, Input, OnChanges, SimpleChanges } from '@angular/core';
-import { PartyMessagesGroup, QueryListData } from '../../domain';
+import { SortOrder } from '../../../complex/sort-order';
+import { PartyMessagesGroup, QueryListData, column } from '../../domain';
 
 @Component({
   selector: 'ccd-query-list',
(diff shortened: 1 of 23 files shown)


projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/components/query-list/query-list.component.ts
```diff
@@ -1,18 +1,90 @@
...
+        this.sort(col);
+        break;
+      }
+    }
+  }
 
-  constructor() {
+  public sortWidget(col: column): string {
+    switch (col.sortOrder) {
+      case SortOrder.DESCENDING: {
```
MateiHMCTS, line 55: I think the order should be different here;

DESCENDING should be a down triangle
ASCENDING should be an up triangle (smaller at the top and increasing as it goes to bottom).

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/components/query-list/query-list.component.ts
```diff
@@ -9,11 +10,81 @@ import { PartyMessagesGroup, QueryListData } from '../../domain';
...
+      }
+      case 'Last response date': {
+        this.sortDate(col);
+        break;
+      }
+      case 'Response by': {
+        this.sort(col);
+        break;
+      }
+    }
```
Daniel-Lam, line 50: @DavidJayakumar - minor point, the linting might pick up that there's no `default` clause... might be worth adding one

Daniel-Lam (changes requested):
@DavidJayakumar - just one minor observation; the rest is OK

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/components/query-list/query-list.component.ts
DavidJayakumar: changed

Daniel-Lam (dismissed):

MateiHMCTS (approved):

PR #1469, opened 2023-06-01
Bug/EUI 8467

DavidJayakumar:

https://tools.hmcts.net/jira/browse/EUI-8467

### Change description ###

Fix the combinations of event and summary

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-edit/case-edit.component.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-edit/case-edit.component.ts
@@ -409,7 +409,7 @@ private caseSubmit({ form, caseEventData, submit }: CaseEditCaseSubmit ): void {
         this.sessionStorageService.removeItem('eventUrl');
         const confirmation: Confirmation = this.buildConfirmation(response);
         if (confirmation && (confirmation.getHeader() || confirmation.getBody())) {
-          this.confirm(confirmation).then().catch();
+          this.confirm(confirmation);
         } else {
           this.emitSubmitted(response);
         }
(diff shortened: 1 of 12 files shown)


MateiHMCTS (changes requested):
I don't understand why there is a need for slice and then two reverses? Wouldn't the two reverses cancel out?

DavidJayakumar:
> I don't understand why there is a need for slice and then two reverses? Wouldn't the two reverses will cancel out?

I took this logic from from [here](https://stackoverflow.com/questions/2218999/how-to-remove-all-duplicates-from-an-array-of-objects) If we are not using this logic, it retains the last object, so the order of the array gets changed and hence tabs are not apppering properly.

Daniel-Lam (changes requested):
@DavidJayakumar - changes look good; would be good if there's a unit test for the `loadCaseDetails()` function in `case-viewer.component.ts`

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-edit/case-edit.component.ts
```diff
@@ -409,7 +409,7 @@ private caseSubmit({ form, caseEventData, submit }: CaseEditCaseSubmit ): void {
         this.sessionStorageService.removeItem('eventUrl');
         const confirmation: Confirmation = this.buildConfirmation(response);
         if (confirmation && (confirmation.getHeader() || confirmation.getBody())) {
-          this.confirm(confirmation);
+          this.confirm(confirmation).then().catch();
```
Daniel-Lam, line 412: @DavidJayakumar - empty `catch()` clause? If intentional, please consider adding a comment to explain

Daniel-Lam (changes requested):
@DavidJayakumar - one other point to look at

MateiHMCTS (approved):

Mo-Lala84 (approved):
approved

Daniel-Lam (approved):

PR #1491, opened 2023-06-15
EUI-8522 QQ create data model and mock data

johnbenjamin-hmcts:
https://tools.hmcts.net/jira/browse/EUI-8522

### Change description ###
Create DataModel and mockdata

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/qualifying-questions/casetype-qualifying-questions.model.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/qualifying-questions/casetype-qualifying-questions.model.ts
@@ -1,6 +1,3 @@
 import { QualifyingQuestion } from './qualifying-question.model';
 
-export interface CaseTypeQualifyingQuestions {
-  caseTypeId: string;
-  qualifyingQuestions: QualifyingQuestion[];
-}
+export type CaseTypeQualifyingQuestions = { [caseTypeId: string]: QualifyingQuestion[] };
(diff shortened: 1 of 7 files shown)


projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/qualifying-questions/casetype-qualifying-questions.model.ts
```diff
@@ -0,0 +1,6 @@
+import { QualifyingQuestion } from './qualifying-question.model';
+
+export interface CaseTypeQualifyingQuestions {
```
ronaldmansveld, line 3: casetype id should be the key for an object. Thus:

```
export type CaseTypeQualifyingQuestions = { [caseTypeId: string]: QualifyingQuestion[] };

ronaldmansveld (changes requested):

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/qualifying-questions/casetype-qualifying-questions.model.ts
ronaldmansveld: Note that your interface would be sub-optimal, as it requires looping through the whole set and trying to match by name. By making it an object where the id is the key, you can immediately refer to it.

johnbenjamin-hmcts: @ronaldmansveld I have now updated the code as you suggested.

ronaldmansveld (approved):

PR #1544, opened 2023-08-16
Bug/fix retries

codaimaster:
https://tools.hmcts.net/jira/browse/EXUI-700

### Change description ###
Fix safe-guard for retry mechanism

```
[ ] Yes
[X] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
@@ -106,7 +106,10 @@ export class CasesService {
     let http$ = this.http.get(url, { headers, observe: 'body' });
 
     const artificialDelay: number = this.appConfig.getTimeoutsCaseRetrievalArtificialDelay();
-    http$ = this.retryUtil.pipeTimeoutMechanismOn(http$, artificialDelay, this.appConfig.getTimeoutsForCaseRetrieval());
+    const timeoutPeriods = this.appConfig.getTimeoutsForCaseRetrieval();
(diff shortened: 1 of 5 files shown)


projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
```diff
@@ -107,8 +107,11 @@ export class CasesService {
 
     const artificialDelay: number = this.appConfig.getTimeoutsCaseRetrievalArtificialDelay();
     const timeoutPeriods = this.appConfig.getTimeoutsForCaseRetrieval();
-    if (timeoutPeriods && timeoutPeriods.length > 0 && timeoutPeriods[1] > 0) {
+    console.log(`Timeout periods: ${timeoutPeriods} seconds.`);
+    if (timeoutPeriods && timeoutPeriods.length > 0 && timeoutPeriods[0] > 0) {
       http$ = this.retryUtil.pipeTimeoutMechanismOn(http$, artificialDelay, timeoutPeriods);
+    } else {
+      console.warn('Skipping to pipe a retry mechanism!');
```
andywilkinshmcts, line 114: Not quite sure what this message means, and I don't think it's really a warning. If we fix the underlying problem we might want to disable the retry. Maybe the message should just be "retry mechanism not enabled" ?

andywilkinshmcts (approved):
Maybe change the console.warn about retry to console.debug

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
codaimaster: Well, we really should keep this retry mechanism for its general purpose. Otherwise, any momentary network glitch will make the application run into the same indefinite spinning wheel issue.

PR #1570, opened 2023-09-27
EUI-8706: Case Flags v2.1 - Determine correct location for adding new flags

Daniel-Lam:
EUI-8706

### Change description ###
Determine the correct location for a flag to be added - either internal or external - based on whether the flag is marked as "internal only" if of type "Other", or its `externallyAvailable` attribute if not of type "Other".

```
[ ] Yes
[x] No
```


connorpgpmcelroy (approved):
LGTM

PR #1599, opened 2023-10-20
initial fix for document upload problems

andywilkinshmcts:

- [x ] commit messages are meaningful and follow good commit message guidelines
- [ x] README and other documentation has been updated / added (if needed)
- [ x] tests have been updated / new tests has been added (if needed)

### Change description ###
Fix to issue caused by cases with many document collections. A bit of research lead me to this:
https://stackoverflow.com/questions/52108773/handling-multiple-components-with-click-events-in-angular
Which points out as an aside that having a hostlistener causes additional change detection cycles. When I checked we do have such a host listener. It doesn't look as though it's adding much value - looks like it it's just setting the error message if the user clicks on the page. 
The same SO post suggests a way of doing this without host listener, which we could try if the functionality is necessary

```
[ ] Yes
[ ] No
```


OgunyemiO (approved):

RiteshHMCTS (changes requested):
@andywilkinshmcts change the version in below file too
projects/ccd-case-ui-toolkit/package.json

RiteshHMCTS (approved):

kasi-subramaniam (approved):

OgunyemiO (approved):

PR #1626, opened 2023-11-29
validation unit test mapArrayValuesToNull

OgunyemiO:

https://tools.hmcts.net/jira/browse/EXUI-1121

### Change description ###

```
[ ] Yes
[x] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/services/fields/fields.purger.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/services/fields/fields.purger.ts
@@ -34,9 +34,9 @@ export class FieldsPurger {
   }
 
   private clearHiddenFieldForFieldShowCondition(currentPageId: string,
-                                                form: FormGroup,
-                                                wizard: Wizard,
-                                                eventTrigger: CaseEventTrigger): void {
+    form: FormGroup,
+    wizard: Wizard,
+    eventTrigger: CaseEventTrigger): void {
     const formFields = form.getRawValue();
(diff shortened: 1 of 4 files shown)


projects/ccd-case-ui-toolkit/src/lib/shared/services/fields/fields.purger.ts
```diff
@@ -309,10 +309,18 @@ export class FieldsPurger {
...
    */
   public mapArrayValuesToNull(array: any[]): any[] {
-    return array.map(element => {
-      return typeof element === 'object'
-        ? Object.assign({}, ...Object.keys(element).map(k => ({ [k]: null })))
-        : null;
-    });
+    if (array && array.length > 0) {
+      return array.map(element => {
+        if (element !== undefined || element === null) {
```
andywilkinshmcts, line 314: Not sure what element === null is adding here. Would work the same without that - or am I missing something

andywilkinshmcts (dismissed):

projects/ccd-case-ui-toolkit/src/lib/shared/services/fields/fields.purger.ts
OgunyemiO: @andywilkinshmcts sometimes the value returned is NULL and not just undefined. In such situation, the code fails and that's why I added that extra check.

RiteshHMCTS (commented):
@OgunyemiO  can you add more unit tests to pass the threshold limit of 80% coverage

OgunyemiO:
The issue raised in this ticket is now covered with the fix in https://tools.hmcts.net/jira/browse/EXUI-1112

PR #1694, opened 2024-03-12
changed serviceId to serviceCode as per requirment

shahedhmcts:

https://tools.hmcts.net/jira/browse/COT-771

### Change description ###
Incorrectly passing service Id. Corrected by passing service code

```
[ ] Yes
[X] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "7.0.21",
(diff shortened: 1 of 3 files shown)


andywilkinshmcts (approved):

andywilkinshmcts (changes requested):
This doesn't work, as we've subsequently discovered

andywilkinshmcts:
I don't think this is an ExUI issue, so this change isn't required

PR #1715, opened 2024-04-12
Task Completion PRL issue EXUI-1793

andywilkinshmcts:

https://tools.hmcts.net/jira/browse/EXUI-1793

### Change description ###

remove isComplete route parameter. Doesn't seem to do anything useful.
Move code that removes the taskToComplete parameter to only do it when the taskToComplete code might then be reeded

```
[ ] Yes
[ x] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/work-allocation.service.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/work-allocation.service.ts
@@ -98,6 +98,7 @@ export class WorkAllocationService {
     if (!this.isWAEnabled()) {
       return of(null);
     }
+    console.log(`completeTask: completing ${taskId}`);
     const url = `${this.appConfig.getWorkAllocationApiUrl()}/task/${taskId}/complete`;
     return this.http
       .post(url, {})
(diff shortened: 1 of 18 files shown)


RiteshHMCTS (approved):

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/work-allocation.service.ts
```diff
@@ -98,6 +102,7 @@ export class WorkAllocationService {
     if (!this.isWAEnabled()) {
       return of(null);
     }
+    console.log(`completeTask: completing ${taskId}`);
```
OgunyemiO, line 105: There are quite a few console logs. Do we really need them?

connorpgpmcelroy (dismissed):
Approving - assuming console.logs for testing purposes

OgunyemiO (dismissed):
Approve with minor comments.

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/work-allocation.service.ts
andywilkinshmcts: I want to leave them in for now. I've actually removed quite a lot that were previously there

PR #1727, opened 2024-05-14
EXUI-1739 - Case link service code fix

Josh-HMCTS:

https://tools.hmcts.net/jira/browse/EXUI-1739?src=confmacro

### Change description ###

Update `getOrgService` to pass in `service_code` to `CaseLinkingReasonCode` api

```
[ ] Yes
[ x ] No
```


connorpgpmcelroy (dismissed):
LGTM

RiteshHMCTS (dismissed):

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/linked-cases/write-linked-cases-field.component.ts
```diff
@@ -92,28 +92,27 @@ export class WriteLinkedCasesFieldComponent extends AbstractFieldWriteComponent
     }
   }
 
-  public getLinkedCaseReasons(serviceId: number): void {
+  public getLinkedCaseReasons(serviceId: number | string): void {
```
andywilkinshmcts, line 95: Does this function actually return a value? I can't see where it does. It should either have a default return, or  should be void.

andywilkinshmcts (changes requested):

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/linked-cases/write-linked-cases-field.component.ts
Josh-HMCTS: The function doesn't return a value so its a void.
I've removed the additional type of 'number' from serviceID as the function is only called in this file and should always be passed as a string

andywilkinshmcts (dismissed):

PR #1816, opened 2024-11-29
update response status

olusegz07:

https://tools.hmcts.net/jira/browse/EXUI-2525

### Change description ###

```
[ ] Yes
[ ] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/query-list/query-list-item/query-list-item.model.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/query-list/query-list-item/query-list-item.model.ts
@@ -50,6 +50,11 @@ export class QueryListItem implements CaseMessage {
   }
 
   public get responseStatus(): QueryItemResponseStatus {
-    return this.children?.length > 0 ? QueryItemResponseStatus.RESPONDED : QueryItemResponseStatus.NEW;
+    if (this.children?.length > 0) {
+      return this.children.length % 2 === 1
+        ? QueryItemResponseStatus.RESPONDED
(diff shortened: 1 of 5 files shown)


connorpgpmcelroy (approved):
LGTM

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/query-management/models/query-list/query-list-item/query-list-item.model.ts
```diff
@@ -50,6 +50,11 @@ export class QueryListItem implements CaseMessage {
   }
 
   public get responseStatus(): QueryItemResponseStatus {
-    return this.children?.length > 0 ? QueryItemResponseStatus.RESPONDED : QueryItemResponseStatus.NEW;
+    if (this.children?.length > 0) {
+      return this.children.length % 2 === 1
```
danlysiak, line 54: Just to check as it doesn't say on the ticket, has this been agreed that the design to determine whether something has been responded to is simply whether there are an odd or even number? Is it ever possible for people to add further clarifications to their own question, for example, which would mess this up?

olusegz07: It has been agreed that users can only respond to or ask follow-up questions one after the other. This means that once a question is asked, the user must wait for a response before they can ask a follow-up question. Similarly, after a response is provided, the user can only ask one follow-up question at a time.

This sequential interaction ensures that the logic remains consistent: the QueryItemResponseStatus.RESPONDED status will always correspond to an odd count, as each response alternates with a follow-up question, forming a predictable sequence.

MunishSharmaHMCTS (approved):

PR #1833, opened 2024-12-18
EXUI-2551: case view not fully expanded

MunishSharmaHMCTS:
https://tools.hmcts.net/jira/browse/EXUI-2551

### Change description ###
case view not fully expended

```
[ ] Yes
[X] No
```


StoneBenHMCTS (dismissed):

andywilkinshmcts (changes requested):
We need to think through and test the implications of this before releasing

andywilkinshmcts (dismissed):
reviewed functionality

PR #1912, opened 2025-04-02
Add the isInternalUser changes

connorpgpmcelroy:
https://tools.hmcts.net/jira/browse/EXUI-2972

### Change description ###
Replace the userIsCaseworker method with a suitable utils function and refactor relevant code

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
+++ b/projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
@@ -1,0 +1,12 @@
+import { SessionStorageService } from "./services";
+
+export const USER_DETAILS = 'userDetails';
+export const PUI_CASE_MANAGER = 'pui-case-manager';
+export const JUDGE = 'judge';
+
+export function isCaseworker(sessionStorageService: SessionStorageService): boolean {
+  const userDetails = JSON.parse(sessionStorageService?.getItem(USER_DETAILS));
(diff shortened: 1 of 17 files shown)


StoneBenHMCTS (approved):

olusegz07 (dismissed):

RiteshHMCTS (approved):

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
```diff
@@ -0,0 +1,12 @@
+import { SessionStorageService } from "./services";
+
+export const USER_DETAILS = 'userDetails';
+export const PUI_CASE_MANAGER = 'pui-case-manager';
+export const JUDGE = 'judge';
+
+export function isCaseworker(sessionStorageService: SessionStorageService): boolean {
```
andywilkinshmcts, line 7: This doesn't quite do what it says on the tin. It might be better to call it isInternalUser, because it returns true for judges too

andywilkinshmcts (changes requested):
I think to avoid future confusion we should rename this to isInternalUser. Bear in mind upcoming changes for other gov departments too.

andywilkinshmcts (dismissed):

RiteshHMCTS (approved):

Josh-HMCTS (approved):

PR #1917, opened 2025-04-04
EXUI-2997 - add reset to event trigger cache

Josh-HMCTS:

https://tools.hmcts.net/jira/browse/EXUI-2997

### Change description ###

add a reset to event trigger cache when leaving an event


andywilkinshmcts (approved):
I like it. Obviously we should do that, now that you've thought about it

StoneBenHMCTS (approved):

anthonydummer (dismissed):

PR #1961, opened 2025-06-25
EXUI-3257 - Simplify cdam exclusion logic

Josh-HMCTS:

https://tools.hmcts.net/jira/browse/EXUI-3257

### Change description ###

Simplify the logic for cdam exclusion as well as document secure mode


connorpgpmcelroy (dismissed):
LGTM

andywilkinshmcts (changes requested):
I think the logic is correct, but we still have 2 bits of logic that essentially do the same thing. I think we need a single function, some thing like isDocumentSecureModeEnabled(appconfig, caseTypeExclusions) 

then in write-document-field we can just do
this.fileSecureMode = isDocumentSecureModeEnabled(appConfig, caseTypeExclusions) 
Then we can get rid of this.secureModeOn, while doesn't seem necessary
In document-management service we can then just do
if (IsDocumentSercureModeEnabled(...) 
 return ..v2
else
return v1

andywilkinshmcts (approved):

PR #2041, opened 2025-11-14
exui-3545-add logs

RiteshHMCTS:
https://tools.hmcts.net/jira/browse/EXUI-3545

### Change description ###
add logs

```
[ ] Yes
[ ] No
```


connorpgpmcelroy (changes requested):
Think caseSubmit (line 457/458) requires logs as well (i.e. log on line 464 before postCompleteTaskIfRequired(), possibly one before this.finishEventCompletionLogic() and one to log if there was an error. I believe this may give us pinpoint info on where the logic is going wrong

PR #2042, opened 2025-11-17
EXUI-3552 - Get case type from URL 

Josh-HMCTS:

https://tools.hmcts.net/jira/browse/EXUI-3552

### Change description ###

There is an edge case where a user can load a casetype through the workbasket, when they go to a case with a different case type and try to upload a file sometimes it will use the wrong casetype for the cdam check, as part of the url updates we have done we can get this value from the url consistently

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,11 @@
 ## RELEASE NOTES
 
+### Version 7.3.1
+**EXUI-3552** Performance issues on screens to return data
+
+### Version 7.3.0
+**COT-775** Angular 20 upgrade
+
 ### Version 7.2.59
 **EXUI-3661** Performance issues on screens to return data
 **EXUI-3681** Users unable to upload documents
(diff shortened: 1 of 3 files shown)


projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -70,20 +70,20 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
 
   public ngOnInit(): void {
     // Wait for both observables to emit at least once
+    const currUrl = window.location.pathname;
+    if (currUrl.includes('/case-details/')) {
+      this.caseTypeId = currUrl.split('/')[4];
```
gregs-cgi: Given this is in a library package, is this URL check robust to all the places we're using it?

Josh-HMCTS: Yes this should cover all cases, this code is based around uploading of documents which is only possible to do when a user is creating a case or in an event. both of which have standard url formats

PR #2065, opened 2025-12-19
EXUI-3892 - Add new code to document management service

Josh-HMCTS:

https://tools.hmcts.net/jira/browse/EXUI-3892

### Change description ###

Update the logic in document management service

```
[ ] Yes
[ ] No
```

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,5 +1,8 @@
 ## RELEASE NOTES
 
+### Version 7.3.13
+**EXUI-3892** Update CDAM logic within document-management.service
+
 ### Version 7.3.12
 **EXUI-2489** Index navigation 2 - media viewer
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hmcts/ccd-case-ui-toolkit",
-  "version": "7.3.10-3892",
+  "version": "7.3.13",
   "engines": {
     "node": ">=20.19.0"
   },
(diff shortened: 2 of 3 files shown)


gregs-cgi (approved):

projects/ccd-case-ui-toolkit/src/lib/shared/components/palette/document/write-document-field.component.ts
```diff
@@ -75,47 +75,41 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
...
+        this.caseTypeId = caseInfo.caseType;
+        this.jurisdictionId = caseInfo.jurisdiction;
+        this.caseId = caseInfo.caseId;
       } else {
-        this.createDocumentForm(document);
+        if (currUrl.includes('/case-details/')) {
+          this.jurisdictionId = currUrl.split('/')[3];
+          this.caseTypeId = currUrl.split('/')[4];
+          this.caseId = currUrl.split('/')[5];
+        }
```
gregs-cgi: Is there a case where we aren't on `case-create`, don't have caseInfo and also aren't on `case-details`? As that would fall through all of these conditionals and not set any data - It could well be that no, that's the only places where this component is used, but it's a fragility here.

gregs-cgi (approved):

RiteshHMCTS (approved):

chrisjones-hmcts (approved):

PR #2127, opened 2026-03-24
exui-4338-case-flag-data-loss

RiteshHMCTS:
https://tools.hmcts.net/jira/browse/EXUI-4338

### Change description ###
case-flag-data-loss

```
[ ] Yes
[x] No
```


Josh-HMCTS (dismissed):
Looks ok to me, one thing i would suggest is that a regression of caseFlags is done as part of this ticket as this effectively skips logic when in caseFlags journey, and i cant remember what functions are actually used in caseFlags journey?

connorpgpmcelroy (dismissed):

PR #2131, opened 2026-03-27
EXUI-4371 - Fix tasks being able to be completed on a different case 

Josh-HMCTS:

https://tools.hmcts.net/jira/browse/EXUI-4371?src=confmacro

### Change description ###

Fix an issue where a task can be incorrectly completed as part of running an event on a different case to the one where the task is available. 
Adds a simple check to compare the current caseId with the caseId on the task.

```
[ ] Yes
[ ] No
```


gregs-cgi (dismissed):

gregs-cgi (dismissed):

connorpgpmcelroy (changes requested):
Looks good and should fix a lot of the problems we are experiencing, however because it is a change related to task completion this should go through SRT to confirm there are no regression issues from, for example, a service that is not expecting this case check.

Question though: what about if the same problem occurs but on the same case e.g. event started with completable task but not finished, event started with possible completable task on same case? Would that issue not still occur? Therefore I would advise checking task type as well

connorpgpmcelroy (dismissed):
As discussed with Josh, this fixes the only issue present - will require SRT though as mentioned

PR #2166, opened 2026-05-18
pofcc-75 - provide a direct mechanism for components to focus on an element outside of the library

jamescatto-cgi:

https://tools.hmcts.net/jira/browse/POFCC-75

### Change description ###

Adds a new service called 'FocusService' that allows components to set the focus on a DOM element that is specific to the service. The 'FocusService' element ID can be used in any DOM element. It is exported as part of the library API to allow wrapper applications to use the element ID - thus embedded components can focus on elements within the wrapper app DOM. By using this service, there is a clear line between the element being focussed and the component that triggers the focussing. 

```
[ ] Yes
[ X] No
```


projects/ccd-case-ui-toolkit/src/lib/shared/services/window/focus.service.ts
```diff
@@ -0,0 +1,20 @@
...
+@Injectable()
+export class FocusService {
+
+    /** unique ID of DOM element this service will focus on */
+    public readonly elementIdToFocus = 'focusService-elementIdToFocus';
+
+    /**
+     * Focus on a specific element with the elementIdToFocus.
+     * If there is no element in the DOM, no action is taken.
+     * TODO Ideally, this service should not be in this library.
```
connorpgpmcelroy: What do you mean by this TODO? If not meant to be in toolkit should it be part of different library or the parent applications?

connorpgpmcelroy (commented):
One comment added

projects/ccd-case-ui-toolkit/src/lib/shared/services/window/focus.service.ts
jamescatto-cgi: I mean if this service was in a separate library, it could be used by other libraries/apps. I just put it here because I didn't know of an appropriate place and AFAIK only this toolkit needs to focus on elements outside of its own components. I can remove the TODO though if you think it is confusing/unnecessary.

connorpgpmcelroy: I would change TODO to NOTE or something of that ilk (or as you've said just remove it) - TODO implies there is work to be done here. This library is already used by a few different applications

jamescatto-cgi: removed it

connorpgpmcelroy (dismissed):

anthonydummer (dismissed):

connorpgpmcelroy (dismissed):

anthonydummer (dismissed):

PR #2190, opened 2026-06-19
EXUI-4216: cgi_ISSUE-035-rpx-xui-manage-organisations-unsafe-eval-csp

chrisjones-hmcts:

[EXUI-4216](https://tools.hmcts.net/jira/browse/EXUI-4216)

### Change description ###

Replaces deprecated PEG.js package with the latest version of its successor peggy and generates the runtime parser from it during the build instead of generating it inline at runtime which is blocked by the MC update to our CSP (remove unsafe eval).

MC PR: https://github.com/hmcts/rpx-xui-webapp/pull/5070

- Added a Peggy grammar source file for the conditional parser.
- Added script to generate condition-parser.generated.ts from the grammar source file.
- Updated condition.peg.ts to import the generated parser instead of generating it inline at runtime.
- Updated the library build so the parser is generated before ng build.
- Moved peggy into devDependencies since it is now only needed at build time.

```
[ ] Yes
[x] No
```


connorpgpmcelroy (dismissed):
Unsure what change involves directly but looks OK to me. Was going to suggest comments in the first file but then said not to edit the code as was a generated file so not an option

olusegz07 (dismissed):

PR #2211, opened 2026-06-29
Bug/ex UI 2645 user by idam redo

connorpgpmcelroy:

https://tools.hmcts.net/jira/browse/EXUI-2645

### Change description ###

Toolkit changes for users by IDAM

```
[ x ] Yes, will need MC to be updated as well
[ ] No
```


projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/case-worker.service.ts
```diff
@@ -16,10 +16,10 @@ export class CaseworkerService {
   ) {
   }
 
-  public getCaseworkers(serviceId): Observable<CaseworkersByService[]> {
-    const url = `${this.appConfig.getWorkAllocationApiUrl()}/caseworker/getUsersByServiceName`;
+  public getUserByIdamId(idamId: string): Observable<Caseworker> {
+    const url = `${this.appConfig.getWorkAllocationApiUrl()}/caseworker/getUserByIdamId`;
     return this.http
-      .post(url, { services: [serviceId]})
+      .post(url, idamId)
```
jamescatto-cgi, line 22: I know you didn't change this, but it is a bit strange to send a POST request to GET a user. If this was a GET request, then the URL could be more RESTFUL, e.g. `/caseworker/{idamId}`

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
```diff
@@ -314,12 +314,15 @@ export class CasesService {
     const userInfoStr = this.sessionStorageService.getItem('userDetails');
     const camUtils = new CaseAccessUtils();
     let userInfo: UserInfo;
-    if (userInfoStr) {
-      userInfo = JSON.parse(userInfoStr);
+    userInfo = userInfoStr ? JSON.parse(userInfoStr) : null;
```
jamescatto-cgi: Maybe not for this PR, but this parsing seems to be done in different places and not always in a typesafe way. I think there is some general code for this in `utils.ts`. It would be nice if we could reuse that and ideally make it typesafe, which also would make it easier to understand that the stored object is not of type `UserDetails` but `UserInfo`, which is confusing.

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-access-utils/index.ts
```diff
@@ -23,34 +23,37 @@ export class CaseAccessUtils {
...
+        if (this.roleOrCategoryExists(CaseAccessUtils.LEGAL_OPERATIONS_ROLE, roleKeywords)) {
+            roleCategoryList.push(CaseAccessUtils.LEGAL_OPERATIONS_ROLE_CATEGORY);
         }
 
+        return roleCategoryList;
     }
 
-    public roleOrCategoryExists(roleKeyword: string, roleCategory: string, roleKeywords: string[], roleCategories: string[]): boolean {
-        const categoryExists = roleCategories.indexOf(roleCategory) > -1;
+    public roleOrCategoryExists(roleKeyword: string,roleKeywords: string[]): boolean {
```
jamescatto-cgi:
```suggestion
    public roleHasKeyword(keyword: string,roleWords: string[]): boolean {
```

projects/ccd-case-ui-toolkit/src/lib/shared/domain/user/user-info.model.ts
```diff
@@ -8,6 +8,5 @@ export interface UserInfo {
   email: string;
   active: boolean;
   roles: string[];
-  roleCategories: string[];
-  roleCategory?: RoleCategory;
+  roleCategories?: RoleCategory[];
```
jamescatto-cgi, line 11: I was wondering about how these 2 were used together, so it makes sense to remove the singular one

jamescatto-cgi (dismissed):
LGTM

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
```diff
@@ -314,12 +314,15 @@ export class CasesService {
...
+    if (!userInfo) {
+      return throwError(() => new Error('User info not found in session storage'));
     }
 
-    const roleCategory: RoleCategory = userInfo.roleCategory || camUtils.getMappedRoleCategory(userInfo.roles, userInfo.roleCategories);
-    const roleName = camUtils.getAMRoleName('challenged', roleCategory);
+    // EXUI-4758 - getMappedRoleCategories no longer returns a single string, checks all roles to get the most likely roleCategory
+    // Unsure whether we should be using mapped role categories any more - should trust the roleCategories from userInfo if they exist
+    const roleCategories: RoleCategory[] = userInfo.roleCategories || camUtils.getMappedRoleCategories(userInfo.roles);
+    const roleName = camUtils.getAMRoleName('challenged', roleCategories[0] as RoleCategory);
```
RiteshHMCTS: @connorpgpmcelroy if roleCategories has blank array then roleCategories[0] becomes undefined, it should be handle.

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/case-worker.service.ts
connorpgpmcelroy: This was mainly because of its previous form and that within MC the current setting is also post for getUserByIdamId and getUsersByIdamIds. The actual request in the backend is a POST which gets cached and then these gets are just acquiring a sub section of that cache so it's less of a simple answer as to whether we should change this. As that is the case, I will keep this as it is for the moment and this can always be addressed in a TODO or a tech debt ticket

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
connorpgpmcelroy: That's a good recommendation, have done it

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-access-utils/index.ts
connorpgpmcelroy: Suggestion fair, have changed

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/services/cases.service.ts
connorpgpmcelroy: Added handling

jamescatto-cgi (dismissed):

RiteshHMCTS (dismissed):

PR #2198, opened 2026-06-23
Task/ex UI 4758 multiple role categories

connorpgpmcelroy:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/EXUI-4758

### Change description ###
Allow use of multiple role categories in toolkit

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[ x ] No
```

PR #2213, opened 2026-06-30
pofcc-156 - added new role category for Enforcement

jamescatto-cgi:

https://tools.hmcts.net/jira/browse/POFCC-156

### Change description ###

Adds a new RoleCategory for Enforcement.

* Refactors RoleCategory into an enum to avoid string duplication
* consolidates role functions into a single location
* create a RoleKeyword enum to capture the idea of embedded names in roles
* Use new enforcement category to ensure users are given the appropriate access

```
[ ] Yes
[ X] No
```


projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-access-utils/index.spec.ts
```diff
@@ -125,7 +13,7 @@ describe('CaseAccessUtils', () => {
                 'dummy',
                 'dummy',
                 'dummy',
-                'ADMIN',
+                RoleCategory.ADMIN,
```
connorpgpmcelroy: Not sure I agree with this, arguably part of the point of this test is to check the fixed string against any potential RoleCategory.ADMIN string changes

projects/ccd-case-ui-toolkit/src/lib/shared/utils.spec.ts
```diff
@@ -1,4 +1,8 @@
+import { AriaDescriber } from '@angular/cdk/a11y';
```
connorpgpmcelroy: Not needed right?

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
```diff
@@ -14,13 +14,97 @@ function getUserDetails(sessionStorageService: SessionStorageService): any {
 
 export function isInternalUser(sessionStorageService: SessionStorageService): boolean {
   const userDetails = getUserDetails(sessionStorageService);
-  return userDetails && userDetails?.roles
-    && !(userDetails.roles.includes(PUI_CASE_MANAGER)
-      || userDetails.roles.some((role: string) => role.toLowerCase().includes(JUDGE)));
+
+  if (!userDetails || !userDetails?.roles) {
+    return false;
+  } else if (RoleCategory.ENFORCEMENT === getMappedRoleCategory(userDetails?.roles, userDetails?.roleCategories)) {
```
connorpgpmcelroy: Only makes sense if enforcement roles will include judges or roles that include 'judge' in the role (i.e. via logic below) and also that it will not include PUI_CASE_MANAGER. Do we expect that to be the case?

As enforcement will also be gathered by caseworkers i don't see why this additional step is needed for that role category particularly and can cause potential problems with task completion for events so unless this is 100% necessary I would remove this if check

connorpgpmcelroy (changes requested):
Have requested some changes. Sonar has also requested some changes which, while i know most of it is not your code, should be fixed when Sonar flags it.

Please check https://github.com/hmcts/ccd-case-ui-toolkit/pull/2198 for reference of multiple role changes that may affect this and also lesser so, https://github.com/hmcts/ccd-case-ui-toolkit/pull/2211/changes, which will also cause merge conflicts. Especially if this would be added in after those changes

projects/ccd-case-ui-toolkit/src/lib/shared/components/case-editor/case-access-utils/index.spec.ts
jamescatto-cgi: I'm happy to revert this, but IMHO I think it is generally better, since we are not testing an enum value here, we are testing the method call. Then, if we used the type instead of strings everywhere, the compiler would flag errors. The benefits are: it saves having to change things if the enum value changes; it shows clearly what the value is in a method that has a load of 'random' strings.

projects/ccd-case-ui-toolkit/src/lib/shared/utils.spec.ts
jamescatto-cgi: good spot, thanks!

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
jamescatto-cgi: Yes, I think you are right, I shouldn't check the roles themselves. AFAIK the role category should always be used in future. I understand that old IDAM roles exist that will require this embedded category as well, so I think to avoid the problem you mention, I shouldn't re-use `getMappedRoleCategory()` since that also checks the role names. So this would become:
```typescript
if (!userDetails || !userDetails?.roles) {
    return false;
} else if (userDetails?.roleCategories?.includes(RoleCategory.ENFORCEMENT)) {
  return true;
} else {
  return !(userDetails.roles.includes(PUI_CASE_MANAGER) || 
    userDetails.roles.some((role: string) => role.toLowerCase().includes(RoleKeyword.JUDGE)))
}
```
What do you think?

anthonydummer (dismissed):
lgtm

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
```diff
@@ -14,13 +14,100 @@ export function getUserDetails(sessionStorageService: SessionStorageService): Us
 
 export function isInternalUser(sessionStorageService: SessionStorageService): boolean {
   const userDetails = getUserDetails(sessionStorageService);
-  return !!userDetails?.roles
-    && !(userDetails.roles.includes(PUI_CASE_MANAGER)
-      || userDetails.roles.some((role: string) => role.toLowerCase().includes(JUDGE)));
+
+  if (!userDetails?.roles) {
+    return false;
+  } else if (userDetails?.roleCategories?.includes(RoleCategory.ENFORCEMENT)) {
```
anthonydummer: I don't think we need to have this explicit reference to ENFORCEMENT here, as I think this will be caught by the fact it's not a PUI_CASE_MANAGER or a judge.

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
```diff
@@ -14,13 +14,100 @@ export function getUserDetails(sessionStorageService: SessionStorageService): Us
...
+  
+  return userDetails?.roles
+      && !userDetails.roles.includes(PUI_CASE_MANAGER)
+      &&
+      (userDetails.roles.includes('caseworker-ia-iacjudge')
+        || userDetails.roles.includes('caseworker-ia-caseofficer')
+        || userDetails.roles.includes('caseworker-ia-admofficer')
+        || userDetails.roles.includes('caseworker-civil')
+        || userDetails.roles.includes('caseworker-privatelaw')
+        || userDetails?.roleCategories?.includes(RoleCategory.ENFORCEMENT));
```
anthonydummer: Can we take out roleCategory.enforcement here. I have raised an exui ticket, https://tools.hmcts.net/jira/browse/EXUI-5169 to cover this fall back function.  It is missing some idam roles for other on-boarded services and it needs to be updated to check against AM roles as is already being done in the exui mc app.

jamescatto-cgi: sure, so the update in EXUI-5169 will include any new enforcement roles?

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
jamescatto-cgi: removed

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
jamescatto-cgi: removed

projects/ccd-case-ui-toolkit/src/lib/shared/utils.ts
jamescatto-cgi: removed

anthonydummer (approved):
lgtm

olusegz07 (approved):

